PR #6, opened 2020-05-02
Adding github release request mapper for ripgrep

creedasaurus:
Kind of a start. Not sure if all the same os/arch mappings are there so I default a missing on to linux, atm.

### DIFF
--- a/ripgrep/releases.js
+++ b/ripgrep/releases.js
@@ -1,0 +1,106 @@
+'use strict';
+
+// this may need customizations between packages
+const osMap = {
+  macos: /\b(mac|darwin|iPhone|iOS|iPad)/i,
+  linux: /\b(linux)/i,
+  win: /\b(win|microsoft|msft)/i,
+  sunos: /\b(sun)/i,
+  aix: /\b(aix)/i
+};
+
+const archMap = {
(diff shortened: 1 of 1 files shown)


ripgrep/releases.js
```diff
@@ -0,0 +1,106 @@
...
+  s390x: /\b(s390x)/i
+};
+
+const fileExtMap = {
+  deb: /\.deb$/i,
+  pkg: /\.pkg$/i,
+  exe: /\.exe$/i,
+  msi: /\.msi$/i,
+  zip: /\.zip$/i,
+  tar: /\.(tar(\.?(gz)?)|tgz)/i,
```
solderjs, line 30: This is too strict for the general case.
There's bz2, gz, xz, and probably others.

solderjs:
Looks good. I'll test it out. Thanks a tonne!

solderjs:
BTW, what does `?!` do in a RegExp?

creedasaurus:
> BTW, what does `?!` do in a RegExp?

Oh yeah, it's a reverse lookup since `x86-64` matches x86 if it happens to catch that first. The reverse lookup checks behind those characters and just makes sure that `x86` is _not_ followed by `-64` or `_64`.

PR #15, opened 2020-05-05
WIP: switch between installed versions

solderjs:
This behaves as expected:

```bash
webi node@12
webi node@14
```

This yields confusing results (I stay on node 14)

```bash
webi node@12
# a few days later
webi node@14
# a few days later
webi node@12
```

This also yields confusing results (I go back to node 12, because it's in my PATH)

```bash
webi node@12
# a few days later
webi node@14
# a few days later
rm -rf ~/.local/opt/node-v14*
```

### DIFF
--- a/golang/install.bash
+++ b/golang/install.bash
@@ -41,8 +41,23 @@ set -u
 # Install go #
 ###################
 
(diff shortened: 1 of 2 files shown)


solderjs:
this was added yesterday or so

PR #17, opened 2020-05-08
WIP: install.bat template, node example

Leghari-K:
Set URL, Parse CSV and breakdown into variables.

### DIFF
--- a/_example/install.bat
+++ b/_example/install.bat
@@ -1,1 +1,50 @@
 rem TODO
+@echo off
+rem get OS version
+for /f "tokens=4-5 delims=. " %%i in ('ver') do set VERSION=%%i.%%j
+if "%version%" == "6.3" set currentOS=win8.1
+if "%version%" == "6.2" set currentOS=win8
+if "%version%" == "6.1" set currentOS=win7
+if "%version%" == "6.0" set currentOS=winVista
+if "%version%" == "10.0" set currentOS=win10
+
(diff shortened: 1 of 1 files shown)


_example/install.bat
```diff
@@ -1 +1,50 @@
...
+rem get OS version
+for /f "tokens=4-5 delims=. " %%i in ('ver') do set VERSION=%%i.%%j
+if "%version%" == "6.3" set currentOS=win8.1
+if "%version%" == "6.2" set currentOS=win8
+if "%version%" == "6.1" set currentOS=win7
+if "%version%" == "6.0" set currentOS=winVista
+if "%version%" == "10.0" set currentOS=win10
+
+rem get this Computer Architecture
+set currentArch=%processor_architecture%
```
jkunkee, line 12: Unfortunately this isn't always this simple. If you're running in a WOW64 process (x86 on x64, x86 on ARM64, ARM32 on ARM64) then there's another check needed here for `PROCESSOR_ARCHITEW6432`:
https://docs.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details

(Note that, for various reasons, PowerShell on ARM64 is x86.)

solderjs:
Mutch of the logic that was to be done here has since moved to the server, and other wis moving into PowerShell via https://github.com/webinstall/packages/pull/35/files

That said, there's still a bit to learn from this.

PR #24, opened 2020-06-07
hugo installer

ryanburnette:
- [x] macOS
- [x] Linux
- [ ] Windows

### DIFF
--- a/hugo/install.bash
+++ b/hugo/install.bash
@@ -1,0 +1,64 @@
+# title: Hugo
+# homepage: https://github.com/gohugoio/hugo
+# tagline: Hugo is a static HTML and CSS website generator written in Go.
+# description: |
+#   Hugo is a static HTML and CSS website generator written in Go. It is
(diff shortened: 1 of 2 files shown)


ryanburnette:
I copied the caddy installer, but hugo works on windows too.

solderjs:
Hugo releases look sane: https://github.com/gohugoio/hugo/releases

Does your Hugo installer work? If not, can you test following the `ripgrep` example instead?

(the only difference is that Caddy does NOT use subfolders where as ripgrep does)

ryanburnette:
- [x] macOS
- [x] Linux
- [ ] Windows

solderjs:
I used the new caddy as a template for this

When you add something new that's a single file type of thing like `rg`, `caddy`, `hugo`, `gitea`, etc, copy from the caddy template and replace `pkg_get_current_version` with a function that properly parses the version.

If it's something more like `node`, `go`, `flutter`, that has a full environment with it, use `node` as a template and likewise replace the `pkg_get_current_version`.

PR #41, opened 2020-07-12
typo fix

solderjs:
subjecting myself to the policy of PR-ing every commit rather than being a cowboy...


ryanburnette:
approved

PR #60, opened 2020-07-27
add bypass alias to bat cheat sheet fix #58

### LINKED ISSUE #58: suggestion for bat cheat sheet, bypass alias
a suggestion to add something to the bat cheat sheet

if you're aliasing cat to bat, it would be nice to teach folks the way to bypass the alias

```bash
\cat foo
```

aaronkchsu:


solderjs:
Note for the future: if you put the issue number in the description, it'll link it to the original issue.

(this doesn't work for titles, apparently)

solderjs:
Ooops! This was supposed to go into the related section "How to alias as cat", not at the bottom in the usage section.

Would you mind fixing that @aaronkchsu ?

PR #62, opened 2020-07-29
add nerdfont

solderjs:
Just another simple thing that I want an easier way to install on new systems (and necessary before adding `lsd`).


solderjs:
Preview at https://beta.webinstall.dev/nerdfont

Install beta:

```bash
export WEBI_HOST=https://beta.webinstall.dev
webi nerdfont
```

Issue #63, opened 2020-07-29
Show the download source when downloading.


solderjs:
Currently we show were a download will be downloaded:

```txt
Downloading node to /Users/aj/Downloads/node-v14.7.0-darwin-x64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 30.2M  100 30.2M    0     0   879k      0  0:00:35  0:00:35 --:--:--  987k
```

To increase transparency and trust, it may be good to show where the download is coming from:

```txt
Downloading https://nodejs.org/dist/v12.18.3/node-v12.18.3-darwin-x64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 30.2M  100 30.2M    0     0   879k      0  0:00:35  0:00:35 --:--:--  987k

Saved node to /Users/aj/Downloads/node-v12.18.3-darwin-x64.tar.gz
```

Two things that need to be solved for here:

- [ ] How can we show (potentially long) URLs in a friendly way?
- [ ] Update the `webi_download` function `_webi/template.sh` with the new wording

## Skills required

- Visual / Verbal Communication
- Basic Command Line knowledge (`mkdir`, `ls`, `tar`, `unzip`)

It's fine for the display to be brainstormed by one person and the implementation to be done by another. However, this is probably basic enough for anyone with minimal bash experience to make the change.

yollotltamayo:
Hello this is my idea for the friendly URLs , show _package_ name and then _download source_ , the URL could be a direct link to the page, also i can implement it , let me know your opinion .
----------------------------------------------------------------------------------------------------------------------------------
 ```
 Downloading node from https://nodejs.org/dist/v12.18.3/node-v12.18.3-darwin-x64.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 30.2M  100 30.2M    0     0   879k      0  0:00:35  0:00:35 --:--:--  987k

Saved node to /Users/aj/Downloads/node-v12.18.3-darwin-x64.tar.gz
```

solderjs:
Sounds good to me.

Here's another thought:

```txt
Downloading node from
https://nodejs.org/dist/v12.18.3/node-v12.18.3-darwin-x64.tar.gz
[...]

PR #64, opened 2020-07-30
agregate the download source #63

yollotltamayo:
This is my first PR. Let me know if you have any feedback!

### DIFF
--- a/_webi/template.sh
+++ b/_webi/template.sh
@@ -147,8 +147,8 @@ webi_download() {
         return 0
     fi
 
-    echo "Downloading $PKG_NAME to $my_dl"
-
+    echo "Downloading $PKG_NAME from:"
+    echo -e "\e]8;;$WEBI_PKG_URL\a$WEBI_PKG_URL\e]8;;\a"
     # It's only 2020, we can't expect to have reliable CLI tools
     # to tell us the size of a file as part of a base system...
     if [ -n "$WEBI_WGET" ]; then
@@ -166,7 +166,8 @@ webi_download() {
         # TODO curl -fsSL --remote-name --remote-header-name --write-out "$my_url"
         curl -fSL -H "User-Agent: curl $WEBI_UA" "$my_url" -o "$my_dl.part"
     fi
-
+    echo ""
+    echo "Saved as $my_dl"
     mv "$my_dl.part" "$my_dl"
 }
 
--- a/_webi/template.sh
+++ b/_webi/template.sh
@@ -147,8 +147,8 @@ webi_download() {
         return 0
     fi
 
-    echo "Downloading $PKG_NAME from:"
-    echo -e "\e]8;;$WEBI_PKG_URL\a$WEBI_PKG_URL\e]8;;\a"
+    echo "Downloading $PKG_NAME from:" #Show the package and the source URL  
+    echo "$WEBI_PKG_URL"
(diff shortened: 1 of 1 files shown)


solderjs:
Interesting. I myself do not know what `\e]8;;` and `\a` are. What are they?

And this may sound like a stupid question... but is that compatible across Terminal.app, iTerm, GnomeTerminal (?), WSL, ssh, bash, zsh, and fish?

And if there's anywhere that it isn't compatible, will it safely be ignored and display _The Right Thing™_ anyway?

If it is overwhelmingly supported across platforms (Mac, Linux, WSL) and architectures, I'd like to have some comments that explain those pieces such that others encountering it understand the incantation. 🧙‍♂️ 

And welcome to to the wide world of Git and Pull and Merge Requests! :)

solderjs:
In my brief testing, it looks like it adds a dashed underline in `fish` and `zsh` on iTerm, but `bash` cannot understand it:

```bash
echo -e "\e]8;;https://example.com\ahttps://example.com\e]8;;\a"
\e]8;;https://example.comhttps://example.com\e]8;;
```

Since `bash` is the most widely used shell, I think it'll be best to drop that. If you'd like to add some other form of decoration that is cross-shell and cross-terminal compatible, I'd be fine with that (again, with comments).

solderjs:
Hmmm... actually, it may just be that `echo -e` doesn't understand it in bash. On a hunch I tried `printf` and I think it's doing the trick:

```
echo $SHELL
/bin/bash

printf "\e]8;;https://example.com\ahttps://example.com\e]8;;\a\n"
https://example.com
```

I also tried it in `Terminal.app` and it is simply ignoring the formatting.

The behavior seems to not change over SSH.

I would imagine that any recent version of WSL should print properly... but that would be good to test.

yollotltamayo:
Hello ,`\e]8;;` and  `\a` are part of a  feature in VTE's based terminals(you can check it out [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) for the full answer) ,also in the same page the apps and terminal that support the incantation :mage_man: are specified .

The problem of showing the link may be solve by the terminal itself , most terminals auto detect the URL , the above feature `'\e]8;;http://example.com\e\\This is a link\e]8;;\e\\\n'` is used for displaying links in specified words, so a simply `echo` will solve the link problem , (I would have liked to know this before , my bad), also thank you for the prompt answer  ,right now i'm finishing the new PR.

solderjs:
Thanks.

Also: `git pull --rebase https://github.com/webinstall/packages master` and then `git push --force orogin download_source` if you don't mind.

solderjs:
Doesn't need comments for the plain echo, just wanted comments if you were going to keep in the `\e]8;;`.

However, not a big. Comments can be left. I'll approve and squash merge.

yollotltamayo:
Just made the new PR ,i think now it is fine.

solderjs:
P.S. Have you checked out https://webinstall.dev?

Lot's of very simple, easy to use developer tools:

- https://webinstall.dev/rg
- https://webinstall.dev/curlie
- https://webinstall.dev/bat
- https://webinstall.dev/myip
- https://webinstall.dev/ssh-pubkey

solderjs:
Thanks for contributing. :)

Issue #69, opened 2020-08-01
correct Prettier cheat sheet about `--ignore`


solderjs:
As seen on https://webinstall.dev/prettier/ the Prettier cheat sheet, the `--ignore` option is incorrect.

- `--ignore` is not the correct argument to pass names of directories to ignore
- need to lookup in prettier docs the correct flag name (if any), or remove

Also, the format of `.prettierignore` is incorrect.

- paths should not begin with `./`, they should be left bare
- `./dist` should be `dist/`

Needs to be edited here: https://github.com/webinstall/packages/blob/master/prettier/README.md

ctyin:
First timer here: could I be assigned this?

solderjs:
Thanks! We'll merge this in today.

PR #70, opened 2020-08-01
correct Prettier cheat sheet about `--ignore`

ctyin:
Addressing issue #69

Closest flag to this is the --ignore-path option in the Prettier docs, but it's different (path to a list of ignored directories). Doesn't seem necessary to include here as it already looks for the ./.prettierignore file


solderjs:
Live at https://webinstall.dev/prettier/

Thanks again @ctyin :)

PR #72, opened 2020-08-04
WIP less magic, more pkg_install

coolaj86:
I think that the general patterns are well-established enough that it makes sense to move some of the installer code back into the `install.sh`

The upside is that the `install.sh` will be (hopefully) be easier to intuit without opening `template.sh`.

The downside is that `install.sh` will have more lines of bash.

- [ ] we don't actually need ` pkg_get_current_version` (we can use the file system install path)
- [ ] the most important and potentially unique commands are in `pkg_install`
    - (the `mv` from the extracted folder to the install path)

### DIFF
--- a/_webi/template.sh
+++ b/_webi/template.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-{
+_webi_run() {
 
 set -e
 set -u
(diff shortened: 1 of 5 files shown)


coolaj86:
I'll come back to this someday, maybe.

PR #85, opened 2020-08-25
Update nerdfont README and support Windows 10

solderjs:
- Correctly add The Nerd Font to Windows 10
- Add instructions for Windows Terminal 1.0

Preview at https://beta.webinstall.dev/nerdfont

@jkunkee Could you take a peek at this? Just want to know if you have a better solution.


solderjs:
@jkunkee Specifically https://github.com/webinstall/packages/pull/85/files#diff-daa776c67381b3ba82c267cf7c15cecc

jkunkee:
`& move` could probably be replaced with `Rename-Item`, but otherwise it's as clean as I could've written it. Hopefully `curl.exe` has stderr behavior that's easier to accommodate in PowerShell than Git...

solderjs:
@jkunkee Thanks for taking a look. I was hoping there might be a less nasty, but equally reliable way of installing fonts.

So far I haven't been dealing with stderr in powershell.

PR #90, opened 2020-08-30
exe added to beginning of formats array

Akshansh99:


solderjs:
Hello Akshansh! It looks like your editor changed every line of the file by converting the indentation.

Would you mind creating a different PR that only changes the one line?

For the future I'd recommend either using VSCode or vim because they both support `Prettier`, which is the formatter that most projects are using these days. We use `Prettier` in this project: https://github.com/webinstall/packages/blob/master/.prettierrc

_webi/test.js
```diff
@@ -4,26 +4,26 @@
 // Print help if there's no pkgdir argument
 //
 var usage = [
-  'Usage: node _webi/test.js --debug <path-to-package>',
-  'Example: node _webi/test.js --debug ./node/'
+    'Usage: node _webi/test.js --debug <path-to-package>',
```
solderjs, line 7: This PR changes almost every line of the file. See my note in the comment section of this PR.

Akshansh99:
I've resolved the issue in PR #91 . Please look into it and thanks @solderjs .

PR #91, opened 2020-08-30
exe added to format array

Akshansh99:
I've correctly formatted the code and avoided any duplication of lines.

### DIFF
--- a/_webi/test.js
+++ b/_webi/test.js
@@ -66,7 +66,7 @@ Releases.get(path.join(process.cwd(), pkgdir)).then(function (all) {
   var pkgname = path.basename(pkgdir.replace(/\/$/, ''));
   var osrel = os.platform() + '-' + os.release();
   var arch = os.arch();
-  var formats = ['xz', 'tar', 'zip'];
+  var formats = ['exe', 'xz', 'tar', 'zip'];
 
   var rel = all.releases.filter(function (rel) {
     return (


solderjs:
That actually doesn't follow the same spacing as the others - again, I'd highly recommend VSCode or `vim` + Prettier. It will make it much more pleasant for you to work with open source projects in the future.

Check out https://webinstall.dev/prettier/

The VSCode plugin is easy to find in their search and I'll be adding http://webinstall.dev/vim-prettier soon https://github.com/webinstall/packages/issues/86

However, this is good enough to accept.

Thanks for contributing.

solderjs:
Oh, I see that you updated the spacing, thanks.

Do you know about `git rebase -i master`? You can `squash` extra commits away so that the merge is very clean and then `git push --force <REMOTE> <BRANCH>` to overwrite the previous commit.

I'll still accept this as-is, but if you'd like to try that, this would be a very simple and perfect opportunity to do so.

Akshansh99:
No I didn't knew about these git commands. I'll do these. Thanks for the guidance!

solderjs:
Hmmm... it looks like that didn't quite work as expected. It should have squashed into just one single commit, but now there are 4 commits:

https://github.com/webinstall/packages/pull/91/commits

Try `git rebase -i 4` instead of `git rebase -i master`. It may be that your master was based on those extra commits.

Akshansh99:
I've squashed all the extra commits into a single commit :

https://github.com/webinstall/packages/pull/91/commits

solderjs:
Perfect. Thanks.

PR #97, opened 2020-08-31
Fix #96: Added required files to the readme.

### LINKED ISSUE #96: [Feat] Add test and temp files to .gitignore
When you run `node _webi/test.js XXXX/` it creates the following:

- install-XXXX.sh
- install-XXXX.ps1

A couple of people have created PRs that have these junk files in them by mistake.

Adding patterns that match (only) these files to `.gitignore` will solve this problem. I think that these patterns should do the trick:
- `install-*.sh`
- `install-*.ps1`

Also, when someone is working in vim it creates a swap files like these:
- `.install.sh.swp`
- `.install.sh.swo` 

We should also add a pattern to match (only) those files.

TanveshT:
As in the issue #96 it was mentioned to add the required patterns to the `.gitignore` file I have did that and also as `install-*.sh` was already present there was no required to add that pattern.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@ node_modules
 install-*.sh
 install-*.bat
 install-*.ps1
-.install.sh.swp
-.install.sh.swo
+.*.sw*
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 node_modules
 install-*.sh
 install-*.bat
+install-*.ps1
+.install.sh.swp
+.install.sh.swo


.gitignore
```diff
@@ -2,3 +2,6 @@
 node_modules
 install-*.sh
 install-*.bat
+install-*.ps1
+.install.sh.swp
```
solderjs, line 6: What we want to match here is vim swap files, which would be in the format of `.*.sw*`

TanveshT: Okay I will test what's the generated files are and accordingly make changes

PR #118, opened 2020-09-16
rg: tweak the wording in the ripgrep description

BurntSushi:
The commit messages provide a tiny bit more detail, but basically:

1. The "rip" in "ripgrep" means "to rip through text" and not "kill grep" or "grep is dead."
2. Dial back some embellishment about ripgrep's performance.

Disclaimer: I'm the author of ripgrep.

Thanks!

### DIFF
--- a/rg/README.md
+++ b/rg/README.md
@@ -16,9 +16,9 @@ Use the `@beta` tag for pre-releases.
 ## Cheat Sheet
 
 > Ripgrep (`rg`) is smart. It's like grep if grep were built for code. It
-> respects `.gitignore` and `.ignore`, has all of the sensible options you want
-> (colors, numbers, etc) turned on by default, is written in Rust, and simply
-> outperforms grep in every imaginable way.
+> respects `.gitignore` and `.ignore`, has all of the sensible options you
+> want (colors, numbers, etc) turned on by default, is written in Rust, and
+> typically outperforms grep in many use cases.
 
 ```bash
 rg <search-term> # searches recursively, ignoring .git, node_modules, etc
--- a/rg/README.md
+++ b/rg/README.md
@@ -18,7 +18,7 @@ Use the `@beta` tag for pre-releases.
 > Ripgrep (`rg`) is smart. It's like grep if grep were built for code. It
 > respects `.gitignore` and `.ignore`, has all of the sensible options you want
 > (colors, numbers, etc) turned on by default, is written in Rust, and simply
-> outperforms grep in every imaginable way. R.I.P. grep.
+> outperforms grep in every imaginable way.
 
 ```bash
 rg <search-term> # searches recursively, ignoring .git, node_modules, etc


ryanburnette (approved):

ryanburnette:
@BurntSushi Thanks for the input. :smile:

coolaj86:
Hey @BurntSushi,

It's exciting to see you in our PRs. 😊 

Although I intended `R.I.P. grep` as a sticky mnemonic, I can see how as the author you might not want to disrespect the shoulders of the giants upon which the new generation of cross-platform unix tool replacements is being built. 

However, `ripgrep` _is_ a superlative product, and I don't want to downplay just how great it is. I don't know of anyone, who having used `rg`, didn't switch immediately (for all but scripting).

What do you think about something like this:

> and simply outshines grep in every imaginable way.

or this?

> and simply more ergonomic than grep in every imaginable way.

My intend was not so much to assert clock cycle performance but rather productivity / how useful it is to a person.

If not, I'll merge this in as is. LMK. Please and Thanks.

BurntSushi:
Hi! Yeah, I totally appreciate that you're excited and want to spread the good word. I think that's awesome and I'm happy to see folks enjoying a tool I built this much. :-) With that said, personally, I would want to avoid the "every imaginable way" verbiage. I get what you're trying to say, but I think some folks might take that literally, and that might give them the wrong idea. You could say something like, "and is generally more ergonomic than grep" or something to that effect.

coolaj86:
> some folks might take that literally

A) They would mean they have no imagination. 😆  \
B) Still wouldn't be incorrect. Can _you_ imagine a way in which it isn't better?

Merging as-is.

BurntSushi:
Sure. grep is POSIX compatible, more portable and more widely available. For that reason, it's generally more ergonomic and reliable to use it instead of ripgrep inside of scripts.

Also, since grep is POSIX compatible, it has full locale support, where as ripgrep just uses UTF-8 with un-tailored Unicode rules. grep for example can do neat things like `echo 'pokémon' | grep 'pok[[=e=]]mon'` in a way that matches your current locale where as ripgrep does not.

Finally, the "smart filtering" that ripgrep does by default is part of what makes it great and attracts a lot of people to the tool, but reasonable people can disagree about whether it's good to have something like that enabled by default. Even though I try to make the smart filtering behavior clear in the docs, people are still surprised by it and do occasionally decide against using ripgrep because of it.

This topic is similar to a FAQ item that I wrote: https://github.com/BurntSushi/ripgrep/blob/master/FAQ.md#posix4ever

coolaj86:
Thanks. I get it now.

Off-topic-ish:

Suddenly I have no idea what POSIX means - as it obviously does not mean what I thought it meant (i.e. portable across operating systems).

And my google-fu is failing me.

I'm finding [tons](https://standards.ieee.org/project/1003_1.html#Standard) [of](https://en.wikipedia.org/wiki/POSIX) [high-level](https://linuxhint.com/posix-standard/) [meta-articles](https://www.oreilly.com/library/view/learning-the-bash/1565923472/apas02.html) about POSIX, but I can't seem to find any of the actual source material with the list of criteria.

Do you happen to know of any good reference material where I could learn what POSIX actually is?

coolaj86:
Where Google failed, Stackoverflow succeeded: https://stackoverflow.com/a/31865755/151312

I'm guessing these are the relevant sections

CLI Utilities
    - http://pubs.opengroup.org/onlinepubs/9699919799/utilities/contents.html
    - https://pubs.opengroup.org/onlinepubs/9699919799/
Regular Expressions
    - https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09

BurntSushi:
POSIX is an interface. The standard reference material is in your man pages, although my system doesn't actually have a POSIX man page for grep (it just has GNU's man page), and to be honest, it's pretty hard to navigate. Otherwise, yeah, the links you found are what I would have given you.

POSIX and portability are closely connected, because when someone says "grep is portable," what they mean is that the _interface_ of grep is portable. My system for example has GNU grep, where as a macOS machine might have a variant of BSD grep. They are two totally different implementations, but they implement a common subset and that common subset is what is specified by POSIX. So as long as all of your `grep` commands stick to the common subset specified by POSIX, then those commands should be able to work in any POSIX-compatible environment, even if the programs implementing the POSIX interface are not the same.

Of course, in practice, programs like GNU grep implement a lot more than what POSIX specifies. When you couple that with the fact that discovering the POSIX subset of grep is not straight-forward or obvious at all, it turns out that in order to write portable programs, you really need to go out of your way to make sure you aren't using any GNU specific flags. (And there are _a lot_ of them.)

The [Wikipedia article on POSIX](https://en.wikipedia.org/wiki/POSIX) is decentish. But I actually don't know of any good resources that really break POSIX down to easily digestible parts. It might be because it's hard to do that. POSIX covers interfaces of many things, including both end user applications _and_ library interfaces. For example, the standard library for C is specified by POSIX. (There are many different implementations of C's standard library.)

POSIX was born in an era where there was a lot more operating system diversity. Nowadays, Linux, macOS and Windows tend to rule the world. Windows is by far the most popular and obvious example of a non-POSIX environment. ("cygwin" is something you can install on Windows that will provide a POSIX environment for you, which is implemented in terms of the Windows API functions. cygwin is, for example, the standard way to get grep on a Windows system.) Since OS diversity is a lot less today than it was, the significance and importance of POSIX can be hard to see. Moreover, over the years, POSIX has somewhat stagnated and OS's have diverged. What this means is that you wind up needing to write a lot of platform specific code anyway if you want to target multiple environments. So POSIX has, IMO, because less relevant over the years, although it serves an important role on which many platforms are based.

Rust was born in this era of less diversity and it shows. For Rust, POSIX moves to an implementation detail, primarily because 1) POSIX alone isn't enough to provide everything that the Rust standard library provides and 2) we want to support Windows and Windows isn't a POSIX system. The Rust standard library provides file system APIs, string handling, concurrency primitives, etc., which all are normally the domain of the POSIX C standard library. But that POSIX interface is tightly coupled to C, so while Rust might use implementations that satisfy the POSIX C standard library interface in its own implementation, it provides its own versions of those portable APIs.

Happy to answer additional questions if you have them!

coolaj86:
Thank you. That's all for now.

Issue #137, opened 2020-09-30
Needs more vim


coolaj86:
- [x] vim-essentials (meta package)
    - [x] vim-leader
    - [x] vim-shell
    - [x] vim-ale
    - [x] vim-spell
    - [x] vim-whitespace (tabwidth, trim trailing) (see https://github.com/webinstall/webi-installers/issues/117)
- [x] [nerdtree](https://github.com/preservim/nerdtree)
    - `map <C-n> :NERDTreeToggle<CR>`
- [ ] [ctrlp](https://github.com/ctrlpvim/ctrlp.vim)
    ```
    let g:ctrlp_show_hidden = 1
    let g:ctrlp_custom_ignore = { 'dir': '\v[\/](\.git|node_modules|rubygems|\.sass-cache|\.vagrant|bower_components|build|tmp|vendor|dist|public.built)$' }
    set wildignore+=.DS_Store
    set wildignore+=*/node_modules/*
    set wildignore+=*/bower_components/*
    ```
- [ ] vim-supertab
- [x] vim-devicons
- [ ] tabnine-vim
- [ ] vim-ack + ripgrep https://www.freecodecamp.org/news/how-to-search-project-wide-vim-ripgrep-ack/
- [x] bugfix: use `~/.vim/plugins` (manual load) instead of `~/.vim/plugin`

### postgres

```bash
mkdir -p ~/.vim/pack/plugins/start
git clone https://github.com/vim-scripts/dbext.vim.git ~/.vim/pack/plugins/start/dbext
git clone https://github.com/lifepillar/pgsql.vim.git ~/.vim/pack/plugins/start/pgsql
```

```
let g:sql_type_default = 'pgsql'
```

Run a command in vim, without a UI, and quit:

```bash
vim -u NONE -c ":GoInstallBinaries" -c q
```

coolaj86:
Leaving supertab for another day.

PR #139, opened 2020-10-01
update golang, vim-go, and watchexec

coolaj86:
- updates how golang tooling is fetched
- make our own watchexec cheat sheet


coolaj86:
TODO: before I merge this I should update the vim command as per what I discovered today: https://github.com/webinstall/webi-installers/issues/137

ryanburnette (approved):

PR #157, opened 2020-11-09
fix deleting entire ~/.local/bin on uninstall

dym-sh:


coolaj86:
For the average Webi user, everything in `~/.local/bin` would have been put there by `webi`.

I would think that the type of person who is already using `~/.local/bin` and can distinguish between which things were installed by webi and which things were installed by something else.

Do you have something else that's putting stuff in there?

dym-sh:
not necessarily, it might be months between an install and revisiting the local bin, at which point it's not hard to forget which of installed things was used by which (3rd party or self-made) script.

either keep track of webi-installed things, use uniquly created directory, or dont suggest deletion of entire directory in `$PATH`

coolaj86:
Sorry for being slow on this. That's cool that some other programs are actually using `~/.local/bin` (like they should!).

I was going to click the edit button to remove `~/.local/opt/webi`, since that actually isn't used, but it looks like you archived the repo and that's disabled.

want to remove that line and I'll merge this in?

Here's what I'm thinking it should look like:


### How to uninstall Webi

`webi` uses standard paths and touches very little.

If you want to remove just the `webi` script, you can do so like this:

```bash
# Mac, Linux
rm -rf ~/.local/bin/webi
```

On Windows you would instead remove `~/.local/bin/webi.cmd` and `~/.local/bin/webi.ps1`.

If you're not using `pathman` for anything else, you can also remove it.

```bash
rm -f ~/.local/opt/pathman-* ~/.local/bin/pathman
```

`pathman` generates `~/.config/envman/PATH.env`, which may be conditionally referenced in any or all of `~/.bashrc`, `~/.profile`, `~/.zshrc`, and `~/.config/fish/config.fish`. You can remove it, but doing so may also remove `~/.local/bin` from your `PATH`.

```bash
rm -f ~/.config/envman/PATH.env
```

### How to uninstall Webi-installed programs

Except where noted otherwise (such as `wsl`) Webi installs everything into `~/.local/bin` and `~/.local/opt`. Some programs, such as `postgres`, also use `~/.local/share` or `~/.config`.

Before removing any of thees directories, you might want to check and see what else is there:

```bash
ls -AF ~/.local/bin/
ls -AF ~/.local/opt/
ls -AF ~/.local/share/
ls -AF ~/.config/
```

If nothing else is using `~/.local`, you can remove it entirely (but probably NOT `~/.config`).

dym-sh:
i unarchived the repo, please re-try what you wanted to change

coolaj86:
@dym-sh take a look now and see what you think

dym-sh:
sure, looks alright :+1:

coolaj86 (approved):

Issue #173, opened 2021-01-07
[Installer] Add Kubectx


nikersch:
Kubectx is a utility to manage and switch between kubectl / Kubernetes contexts.

Repo: https://github.com/ahmetb/kubectx

A colleague of me will add this.

michaelpaesold:
As Nico said, I added a PR for this. Hope it's fine.

PR #175, opened 2021-01-08
Added kubectx installer

michaelpaesold:
See https://github.com/ahmetb/kubectx

Implements issue #173

Collaborative work of Nico Kerschbaumer and me

### DIFF
--- a/kubectx/README.md
+++ b/kubectx/README.md
@@ -1,0 +1,38 @@
+---
+title: Kubectx
+homepage: https://github.com/ahmetb/kubectx
+tagline: |
+  kubectx: kubectx is a utility to manage and switch between kubectl contexts.
+---
+
+### Updating `kubectx`
+
+`webi kubectx@stable`
+
+Use the `@beta` tag for pre-releases.
+
+## Cheat Sheet
+
+> `kubectx` kubectx helps you switch between Kubernetes clusters back and forth
+
+To run kubectx:
+
+```bash
+kubectx
+```
+
+### Command line arguments
+
+```bash
+USAGE:
+  kubectx                   : list the contexts
+  kubectx <NAME>            : switch to context <NAME>
+  kubectx -                 : switch to the previous context
+  kubectx -c, --current     : show the current context name
+  kubectx <NEW_NAME>=<NAME> : rename context <NAME> to <NEW_NAME>
+  kubectx <NEW_NAME>=.      : rename current-context to <NEW_NAME>
+  kubectx -d <NAME>         : delete context <NAME> ('.' for current-context)
+                              (this command won't delete the user/cluster entry
+                              that is used by the context)
+  kubectx -u, --unset       : unset the current context
+```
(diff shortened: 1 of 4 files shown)


coolaj86:
I'm looking at this now.

coolaj86 (commented):
Looks good enough to add and get started, though I would like some more info, which I'll mention in a moment.

coolaj86 (approved):

coolaj86:
Should this also install `kubens`?

Also, I've made a PR to kubectx to set the version: https://github.com/ahmetb/kubectx/pull/279 Re: https://github.com/ahmetb/kubectx/issues/125

coolaj86:
It would be nice to have a little more info in the Cheat Sheet. I'm not a k8s user, but what's there isn't quite enough for me to get a handle on why I'd use this if I were.

I'd recommend adding some info about why it's personally important to use, or a few quick commands for things that you do on a daily (or otherwise regular) basis.

michaelpaesold:
> Should this also install `kubens`?

Actually yes, should this be a separate installation or is there precedence for bundling the two tools?

coolaj86:
In this case it probably should be a separate installer because it's published as a separate binary:

https://github.com/ahmetb/kubectx/releases/tag/v0.9.1

Otherwise, if it were in the same tarball, I'd give an example of modifying the installer to move both around.

I'd recommend modifying the `releases.js` of both to make sure that `kubectx` is filtered from `kubens` and vice-versa, that might look like this: https://github.com/webinstall/webi-installers/blob/master/caddy/releases.js

```js
    // remove kubens, etc from kubectx list
    all.releases = all.releases.filter(function (rel) {
      return !/kubectx/i.test(rel.name);
    });
```

coolaj86:
Also, would you make a PR to the README or install docs page of `kubectx` to add as well?

For many projects updating the install section might look something like this:

~~~md
## macOS

Homebrew:

```bash
brew install xxxxx
```

Webi:

```bash
curl -sS https://webinstall.dev/xxxxx | bash
```

## Linux

Apt:

```bash
sudo apt -y install xxxxx
```

Webi:

```bash
curl -sS https://webinstall.dev/xxxxx | bash
```

## Windows

Choco:

```bash
choco install xxxxx
```

Webi:

```bash
curl.exe -A MS https://webinstall.dev/xxxxx | powershell
```
~~~

nikersch:
> In this case it probably should be a separate installer because it's published as a separate binary:
> 
> https://github.com/ahmetb/kubectx/releases/tag/v0.9.1
> 
> (otherwise, if it were in the same tarball, I'd give an example of modifying the installer to move both around)
> 
> I'd recommend modifying the `releases.js` of both to make sure that `kubectx` is filtered from `kubens` and vice-versa. That might look like this: https://github.com/webinstall/webi-installers/blob/master/caddy/releases.js
> 
> ```js
>     // remove kubens, etc from kubectx list
>     all.releases = all.releases.filter(function (rel) {
>       return !/kubectx/i.test(rel.name);
>     });
> ```

I can submit a PR for separate kubens installer.

PR #218, opened 2021-04-05
Ps1 cleanup

ishanpan:
Edited files according to specification

### DIFF
--- a/awless/install.ps1
+++ b/awless/install.ps1
@@ -44,14 +44,14 @@ IF (!(Test-Path -Path "$pkg_src_cmd")) {
 
     # Settle unpacked archive into place
     echo "Install Location: $pkg_src_cmd"
-    New-Item "$pkg_src_bin" -ItemType Directory -Force
+    
     Move-Item -Path ".\awless.exe" -Destination "$pkg_src_bin"
 
     # Exit tmp
     popd
 }
-
+New-Item "$pkg_src_bin" -ItemType Directory -Force | out-null
 echo "Copying into '$pkg_dst_cmd' from '$pkg_src_cmd'"
-Remove-Item -Path "$pkg_dst_cmd" -Recurse -ErrorAction Ignore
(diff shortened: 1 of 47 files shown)


awless/install.ps1
```diff
@@ -44,14 +44,14 @@ IF (!(Test-Path -Path "$pkg_src_cmd")) {
 
     # Settle unpacked archive into place
     echo "Install Location: $pkg_src_cmd"
-    New-Item "$pkg_src_bin" -ItemType Directory -Force
```
coolaj86, line 47: this should not be moved, only the instances that are directly nested in an if statement should be moved.

vim-ale/install.ps1
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env pwsh
 
 IF (!(Test-Path -Path "$Env:USERPROFILE\.vim\pack\plugins\start")) {
```
coolaj86, line 3: this is the sort of IF block that should be removed

coolaj86:
Hey @ishanpan, this is not quite right. Please take a more careful look at the description.

I'd recommend doing this in three commits:
1. use an automated tool to add ` | out-null` to the end of every `Remove-Item`, commit
2. use an automated tool to add ` | out-null` to the end of every `New-Item`, commit
3. manually (or automatically) remove all `IF ... Test-Path` blocks, but leave the `New-Item` line in-place (minus the leading whitespace), commit

ishanpan:
> Hey @ishanpan, this is not quite right. Please take a more careful look at the description.
> 
> I'd recommend doing this in three commits:
> 
> 1. use an automated tool to add ` | out-null` to the end of every `Remove-Item`, commit
> 2. use an automated tool to add ` | out-null` to the end of every `New-Item`, commit
> 3. manually (or automatically) remove all `IF ... Test-Path` blocks, but leave the `New-Item` line in-place (minus the leading whitespace), commit

Thank you for the feedback. I will rectify my mistake and get back to you.

PR #219, opened 2021-04-06
cleanup ps1

ishanpan:
I think I have corrected my mistake. Let me know if there's any more issues

### DIFF
--- a/comrak/install.ps1
+++ b/comrak/install.ps1
@@ -36,5 +36,5 @@ IF (!(Test-Path -Path "$Env:USERPROFILE\.local\opt\$Env:PKG_NAME-v$Env:WEBI_VERS
 }
 
 echo "Copying into '$Env:USERPROFILE\.local\bin\$EXENAME' from '$Env:USERPROFILE\.local\opt\$Env:PKG_NAME-v$Env:WEBI_VERSION\bin\$VERNAME'"
-Remove-Item -Path "$Env:USERPROFILE\.local\bin\$EXENAME" -Recurse -ErrorAction Ignore
+Remove-Item -Path "$Env:USERPROFILE\.local\bin\$EXENAME" -Recurse -ErrorAction Ignore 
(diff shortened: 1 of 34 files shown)


coolaj86:
@ishanpan The only problem I see here is that you have removed ALL `IF ... Test-Path` statements, but only the ones directly surrounding `New-Item` should have been removed.

That's only in `_webi/template.ps1` and `_webi/webi-pwsh.ps1`.

comrak/install.ps1
```diff
@@ -36,5 +36,5 @@ IF (!(Test-Path -Path "$Env:USERPROFILE\.local\opt\$Env:PKG_NAME-v$Env:WEBI_VERS
 }
 
 echo "Copying into '$Env:USERPROFILE\.local\bin\$EXENAME' from '$Env:USERPROFILE\.local\opt\$Env:PKG_NAME-v$Env:WEBI_VERSION\bin\$VERNAME'"
-Remove-Item -Path "$Env:USERPROFILE\.local\bin\$EXENAME" -Recurse -ErrorAction Ignore
+Remove-Item -Path "$Env:USERPROFILE\.local\bin\$EXENAME" -Recurse -ErrorAction Ignore
```
coolaj86, line 39: Looks like this is adding an extra space at the end of the line?

dotenv-linter/install.ps1
```diff
@@ -26,8 +26,7 @@ IF (!(Test-Path -Path "$Env:USERPROFILE\Downloads\$Env:WEBI_PKG_FILE"))
     & move "$pkg_download.part" "$pkg_download"
 }
 
-IF (!(Test-Path -Path "$pkg_src_cmd"))
```
coolaj86, line 29: Only the IF blocks directly surrounding New-Item commands should have been removed, not ALL IF blocks.

coolaj86 (changes requested):

PR #254, opened 2021-06-14
Fix breaking newline

PMilla-MS:
New line causes the command to throw an exception instead of correctly installing font


coolaj86 (approved):
Thank you!

PR #273, opened 2021-07-25
[wip] setcap-netbind and vps-essentials

ryanburnette:
- [x] setcap-netbind
- [ ] recreate vps-utils as vps-essentials leaving vps-utils garbage intact
- [ ] follow meta installer pattern from vps-essentials

### DIFF
--- a/setcap-netbind/README.md
+++ b/setcap-netbind/README.md
@@ -1,0 +1,23 @@
+---
+title: setcap-netbind
+homepage: https://github.com/webinstall/webi-installers/setcap-netbind/README.md
+tagline: |
+  setcap-netbind: Give a binary the ability to bind to privileged ports.
+---
+
+setcap-netbind will grant the specified program the ability to listen on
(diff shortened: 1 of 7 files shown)


ryanburnette:
@coolaj86 It looks like the purpose of vps-utils is to bundle multiple installers into one, but I think that's not worth doing. There's a better way if we want to have a bundle of installers that won't involve duplication. I'm seeing myip and vps-myip that look like they do the same thing.

If someone wants vps-utils maybe they should just run this command:

```bash
webi myip setcap-netbind addswap
```

Or we could have a way to set up a bundle by having a README.md with front matter listing the packages and the documentation just links to the included installers.

ryanburnette:
It looks like vim-essentials is a good example of a meta installer. I can make vps-utils match that pattern.

ryanburnette:
Will rename to vps-essentials.

coolaj86 (approved):
I'm taking your WIP for `setcap-netbind` and rolling with it. We can do the full `vps-essentials` meta package later.

PR #328, opened 2021-11-15
bugfix(yq): update installer for new package format of new versions

coolaj86:
The published builds have both raw binaries, as well as man files in the tar, which have `yq` in the name.


coolaj86:
Superseded by https://github.com/webinstall/webi-installers/pull/332

PR #331, opened 2021-11-17
bugfix: shfmt/install.ps1 #314

### LINKED ISSUE #314: problem with shfmt installer
When I tried to install `shfmt`, I got the following errors.

```
Installing shfmt
pushfmtd : The term 'pushfmtd' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At C:\Users\Liu.D.H\.local\tmp\shfmt.install.ps1:71 char:5
+     pushfmtd .local\tmp
+     ~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (pushfmtd:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : CommandNotFoundException

PS C:\Users\Liu.D.H>
```

It seemed the install script had some mistake on https://github.com/webinstall/webi-installers/blob/22d3602c7d426200e3efe244e27bf15eaf6425da/shfmt/install.ps1#L29.

And the download file is an exe binary, not an achieved file. So the following failed too.
https://github.com/webinstall/webi-installers/blob/22d3602c7d426200e3efe244e27bf15eaf6425da/shfmt/install.ps1#L34-L35

liudonghua123:


coolaj86 (approved):
Couldn't boot up my VM to test, but it looks right, so I'm approving it anyway.

It certainly can't break it worse, so there's that... :)

coolaj86:
Confirmed working by Null from the live stream.

Thanks!

Issue #325, opened 2021-11-15
[Internal] Cleanup & Style


coolaj86:
These are tasks that anyone could help with.

- [x] #393
- [x] #394
- [x] `shfmt` all files
   ```bash
   fd -e sh -0 | xargs -0 shfmt -w -i 4 -sr -ci -s
   ```
- [x] Lint all *.sh files (with shellcheck & shfmt). Most install.sh files should disable unexported var warnings
   ```bash
   # shellcheck disable=SC2034
   #  "'pkg_cmd_name' appears unused. Verify it or export it."
   ```
- [x] All bash scripts should begin with `#!/bin/bash`
- [x] All functions in the `*/*.sh` files should be explicitly declared as functions
   ```bash
   # ✅ correct
   function pkg_install() {
       # ...
   }
   ```
   ```bash
   # ❌ incorrect
   pkg_install() {
       # ...
   }
   ```
- [x] All installers should have a wrapper function
   ```bash
   # ✅ correct
   #!/bin/bash
   set -e
   set -u
   
   function __install_foobar() {
       # ...
   }
   __install_foobar
   ```
   ```bash
   # ❌ incorrect
   #!/bin/bash
   set -e
   set -u
   {
       # ...
   }
   ```
- [x] ~~All conditionals should be in the format `if [[ expr ]]; then`, not `if [ expr ]; then`~~ (reverse that, we are POSIX!)
- [x] #395
- [ ] #660
   ```txt
   ~/.bashrc
   ~/.config/envman/PATH.env
   ~/.local/bin/rg
   ~/.local/opt/rg/
   ~/.local/opt/rg-{version}/
   ```
- [ ] Move `pkg_get_current_version` command to cheat sheet

y0rune:
Hi!
I created the https://github.com/webinstall/webi-installers/pull/330. Please check if my changes are correct. 
Thank you ;)

PR #342, opened 2021-11-21
added shebang, set outside of function, added function name

JKeerl:
According to  [Internal] Cleanup & Style #325 
I added shebang, function name and brought set -e and set -u outside of the function.


coolaj86 (approved):
This was missing the part at the end where it actually calls the function, but I added that. Thanks for contributing.

PR #343, opened 2021-11-21
add shebang, set bash strict mode, create function

JKeerl:
Added shebang, set bash strict mode, create function where it was missing. This was done in an automated way, I checked all the files and it seems in order.
According to  [Internal] Cleanup & Style #325 

If something is not right, please let me know!

### DIFF
--- a/arc/install.sh
+++ b/arc/install.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
+set -e
+set -u
+
 function __init_arc() {
-    set -e
-    set -u
 
     ####################
     # Install archiver #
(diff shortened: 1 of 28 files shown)


coolaj86:
Someone else also worked on this (https://github.com/y0rune/webi-installers/commit/2fe42680da57cdefd9df45787e42f705732e291a), so there are some conflicts. 

I'll try to get back to you tomorrow.

coolaj86:
This item would be a great one to work on that no one else is working on:

- [ ] Move all "shmatter" (bash-frontmatter) to README.md frontmatter

Also, when changing many files at once, it's most excellent if you can script the change and show that too - because then it's easy to identify things that would get lost by tired eyes. :)

coolaj86 (approved):
I went through the conflicts and applied the remaining changes.

Thanks!

:)

PR #346, opened 2021-11-21
Add delta

devPalacio:
#320 

Got it working on windows 10 powershell and WSL. Had a strange issue where sending the useragent would brick the download at 1kb, so i removed it from the curl.exe command.

Cant get test.js to work on my M1 mac. Breaks on line 84 of test.js with
`TypeError: Cannot set properties of undefined (setting 'oses')`

Please review, first time attempting this and open to any and all feedback.

### DIFF
--- a/_example/install.sh
+++ b/_example/install.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# shellcheck disable=SC2034
+# "'pkg_cmd_name' appears unused. Verify it or export it."
+
 function __init_foobar() {
     set -e
     set -u
@@ -21,10 +24,10 @@ function __init_foobar() {
     # pkg_install must be defined by every package
     pkg_install() {
         # ~/.local/opt/foobar-v0.99.9/bin
-        mkdir -p "$(dirname $pkg_src_cmd)"
+        mkdir -p "$(dirname "${pkg_src_cmd}")"
 
         # mv ./foobar-*/foo ~/.local/opt/foobar-v0.99.9/bin/foo
-        mv ./foobar-*/foo "$pkg_src_cmd"
+        mv ./foobar-*/foo "${pkg_src_cmd}"
     }
 
     # pkg_get_current_version is recommended, but not required
--- a/_webi/template.ps1
+++ b/_webi/template.ps1
@@ -31,6 +31,11 @@ New-Item -Path .local\opt -ItemType Directory -Force | out-null
 
 function webi_add_path
 {
+    Write-Host ''
+    Write-Host '*****************************' -ForegroundColor red -BackgroundColor white
+    Write-Host '*    IMPORTANT - READ ME    *' -ForegroundColor red -BackgroundColor white
+    Write-Host '*****************************' -ForegroundColor red -BackgroundColor white
+    Write-Host ''
     & "$Env:USERPROFILE\.local\bin\pathman.exe" add "$args[0]"
     # Note: not all of these work as expected, so we use the unix-style, which is most consistent
     #& "$Env:USERPROFILE\.local\bin\pathman.exe" add ~/.local/bin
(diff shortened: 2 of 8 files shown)


coolaj86:
> Cant get test.js to work on my M1 mac.

Hmm... I suppose I never updated the tests to have the same fallback-to-x64 behavior for M1 that the main installer does. I'll have to add a TODO for that.

> Had a strange issue where sending the useragent would brick the download at 1kb.

That is strange. I'm going to try on a Windows VM and see if I can reproduce. It's actually important that the user agent is sent because that let's us distinguish between x32, x64, arm64, and Windows 10 and 11.

**Could you run this** in PowerShell on your computer to help me debug?

```pwsh
# This is the canonical CPU arch when the process is emulated
$my_arch = "$Env:PROCESSOR_ARCHITEW6432"
IF ($my_arch -eq $null -or $my_arch -eq "") {
  # This is the canonical CPU arch when the process is native
  $my_arch = "$Env:PROCESSOR_ARCHITECTURE"
}
echo my_arch: $my_arch

# TODO API should know to prefer x86 for windows when arm binary is not available
$Env:WEBI_UA = "Windows/10 $my_arch"

echo WEBI_UA: $Env:WEBI_UA
```

> first time attempting this

Well you seem to be doing a bang-up job and checking all the right boxes! I would have thought you've contributed before.

Two things:
1. I've been trying to move over to "semantic commit messages" as are many other projects these days. The format generally looks like this:
   ```txt
   feature(delta): add *nix and Windows installer + cheat sheet
   ```
2. It's a good practice to "roll-up" (technically called _fixup_ or _squash_) your commits so that there are only 1 or 2 meaningful line items rather than many with "WIP" or "cleanup" commits. To do that you _interactive_ rebase against `main` (or `master`) and follow the instructions to place `f` in front of commits you want to roll-up into the previous commit, and `r` in front of commits you want to reword:
   ```bash
   git checkout main
   git pull https://github.com/webinstall/webi-installers main
   git checkout add-delta
   git rebase -i main
   ```
   For example:
   ```txt
   reword 48ed7d4 find-replace on delta v1
   fixup 357d6f3 fix ps1 and repo owner
   f b7c9172 Update README.md
   f 0a69e21 cleanup
   f 5511f36 cleanup version on install
   pick 21550cd match new style and add to test

   # Rebase bc91c98..21550cd onto bc91c98 (6 commands)
   #
   # Commands:
   # p, pick = use commit
   # r, reword = use commit, but edit the commit message
   # f, fixup = like "squash", but discard this commit's log message
   ```
   ```bash
   git push --force --set-upstream origin add-delta
   ```

Or, if for some reason things don't apply well, you go back to the kludge of `git reset`:

```bash
git rebase --abort
# find the commit hash just before your first commit
git log
git reset xxxxxx
git stash
git rebase main
git stash apply
# handle conflicts as per usual (but without replaying any history)
```

That said, I went ahead and did some `git rebase`-ing and pushed to your branch already.

Still setting up the Windows VM to test the PS1 script...

coolaj86 (approved):
You did great! and I put in a bit of my own flair, as well as a few other general Webi improvements I found as I was testing.

coolaj86:
P.S. @devPalacio Will you please test https://webinstall.dev/delta on your Windows install, and report back about the `echo $my_arch` mentioned above?

I still want to debug that if it's an issue.

devPalacio:
the code you asked me to run returned
`echo WEBI_UA: $Env:WEBI_UA
my_arch:
AMD64
WEBI_UA:
Windows/10 AMD64`

The install from the website worked without a hitch.

Thank you for all the feedback!

devPalacio (commented):
Great change on running the pathman command, I totally missed that 👍

dxkf177 (commented):
[Uploading Magisk-27.0.zip…]()

Issue #221, opened 2021-04-09
[Windows] ripgrep installer should be an alias of rg


ajeetdsouza:
## What were you trying to install (or what else went wrong)?

ripgrep

## What exactly did you do?

```sh
curl.exe -A "MS" https://webinstall.dev/ripgrep | powershell
```

## What went wrong?

The installer download failed.

```
Downloading https://webinstall.dev/api/installers/ripgrep.ps1?formats=zip,exe,tar
curl: (22) The requested URL returned error: 404 Not Found
.\.local\tmp\ripgrep.install.ps1 : The term '.\.local\tmp\ripgrep.install.ps1' is not recognized as the name of a cmdlet, function,
script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At line:1 char:1
+ .\.local\tmp\ripgrep.install.ps1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (.\.local\tmp\ripgrep.install.ps1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

## Which OS did you try on?

- [ ] Mac
- [x] Windows
- [ ] Linux

PowerShell 7.1.1

## What type of computer (i.e. laptop, desktop, Raspberry Pi)?

- [x] Laptop or Desktop (amd64 / x86-64)
- [ ] Raspberry Pi (ARM-64)
- [ ] Other (please specify, if you know)

coolaj86:
It's just missing the `rg` alias. Do this for the time being:

```pwsh
curl.exe -A "MS" https://webinstall.dev/rg | powershell
```

coolaj86:
# Easy: copy, paste => find & replace

I'm opening this up for anyone to take this on and make the PR.

0. Comment to let me know which of these you want to take on \
   (and then just do it - no need to wait for me to assign)
1. Copy: https://github.com/webinstall/webi-installers/blob/main/git-gpg-init/install.ps1
2. Paste: https://github.com/webinstall/webi-installers/blob/main/ripgrep/install.ps1
3. Find: `git-gpg-init`
4. Replace with `ripgrep`

Want to do a few more? All of these need the same thing:
- [ ] ripgrep
- [ ] go
- [ ] gpg
- [ ] nerd-font
- [ ] nerd-fonts
- [ ] nerdfonts

ajeetdsouza:
[...]

PR #359, opened 2021-11-24
ripgrep / rg alias (see #221)

josuer08:
Making ripgrep installer an alias for rg


kuvaldini:
Would you make the same for Linux,Mac,BSD?

coolaj86:
@kuvaldini the bash installer alias is already there for *nix.

Issue #360, opened 2021-11-26
[Bug] installing k9s for windows fails


koddistortion:
## What were you trying to install (or what else went wrong)?

I was trying to install **k9s** for Windows (x64 Desktop device)

## What exactly did you do?

As stated in the instructions of the [README.md from k9s](https://github.com/derailed/k9s#installation), I ran `curl.exe -A MS https://webinstall.dev/k9s | powershell`

## What went wrong?

After installing the necessary tools, downloading of k9s release fails:
`curl: (6) Could not resolve host: Windows`

## Which OS did you try on?
- [ ] Mac
- [x] Windows
- [ ] Linux

## What type of computer (i.e. laptop, desktop, Raspberry Pi)?
- [x] Laptop or Desktop (amd64 / x86-64)
- [ ] Raspberry Pi (ARM-64)
- [ ] Other (please specify, if you know)

koddistortion:
I created a pull request, see https://github.com/webinstall/webi-installers/pull/361

coolaj86:
Merged. Thank you @koddistortion!

PR #361, opened 2021-11-26
Fixing windows installer curl call (l#360)

koddistortion:
The environment variable $Env:WEBI_UA seems to be set to "Windows " followed by Arch (x64 etc.). Unfortunately, the parameters for the curl.exe call are "misplaced" and the user agent string is not passed to the -A parameter. This commit fixes this issue.

### DIFF
--- a/_example/README.md
+++ b/_example/README.md
@@ -15,6 +15,17 @@ tagline: |
 To update or switch versions, run `webi example@stable` (or `@v2`, `@beta`,
 etc).
 
+### Files
+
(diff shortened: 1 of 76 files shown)


koddistortion:
See the corresponding issue https://github.com/webinstall/webi-installers/issues/360

coolaj86:
Thanks. Looks like a typo snuck in. I'll plan to do some PR reviews tomorrow and I'll pull this in then.

coolaj86 (approved):

PR #362, opened 2021-11-30
Add make-workflows.sh

kuvaldini:
Hello, guys maintainers. There is one simple but useful repo for github actions users, who wants to be DRY and keep going with GHA. 

I like webi idea. It is strange to have just 168 stars. It would be nice to do not  make people to maintain yet another package manager. For example i can install it simply with npm without publishing to anywhere, just having  npmignore and package.json files in repo/. 

I do not like you use several languages and files for installer. releases.js and install.sh could be merged in one file. And since there are many utils with very similar instructions, they would look nicer just saying `webi kuvaldini/make-workflows.sh` to download and install the latest release for GH. Without publishing it here. What do you think about this?

### DIFF
--- a/make-workflows.sh/README.md
+++ b/make-workflows.sh/README.md
@@ -4,3 +4,61 @@ homepage: https://github.com/kuvaldini/make-workflows.sh
 tagline: |
   make-workflows.sh: A tool to support YAML anchors for GitHub Actions Workflows – DRY for GHA 
 ---
+
+To update or switch versions, run `webi make-workflows.sh@stable` (or `@1.0.0`, `@beta`, etc).
+
+
+## Cheat Sheet
+
+> GitHub Workflow description in YAML does not support anchors.
+There are several workarounds => anyway [they](#links) come to building-editing workflow yaml from source.
+> So I suggest yet another one `make-workflows.sh` based on YAML tool [`yq`](https://github.com/mikefarah/yq) version 4.
+
+This simple script expands YAML anchors in GitHub workflows sources.
+
+
+### USAGE
+```
+$ ./make-workflows.sh --help
+make-workflows:
+   This script expands '*.src.yml' from $1..$[N-1] (default: REPO_ROOT/.github/)
+   to $N (default:REPO_ROOT/.github/workflows/) with corresponding name '*.yml'
+   Main goal is to dereference YAML anchors.
+   Deals only with Git cached/indexed files until --worktree passed.
+   DEBUG: use option -x
+   NOTE: spaces in filenames are not allowed to keep code simplicity.
+Usage:
+    make-workflows.sh [--worktree] [dirs_from... [dir_to]]
+    make-workflows.sh [--help]
+Options:
+   --worktree       List files and get contents from working tree
+                    instead of git index
+   -h, --help       show this help
+   -x, --trace, +x, --no-trace   enable/disable bash trace
+   -i, --install
+   --update
+   -V, --version
+```
+
+### Automate using pre-commit (recommended)
+There is a nice tool [pre-commit](https://pre-commit.com) to do checks and some actions just before commit. The tool is called by Git pre-commit hook.
+
+Making workflows is better being automated – just 
+```sh
+$ pre-commit install
+```
+and add next sample to [`.pre-commit-config.yaml`](/.pre-commit-config.yaml)
+```yaml
+repos:
+- repo: local
+  hooks:
+  - id: make-workflows
+    name: Make GitHub workflows from *.src.yml
+    entry: bash -c '.github/make-workflows.sh && git add .github/workflows'
+    language: system
+    files: '.github/.*\.src\.ya?ml'
+    pass_filenames: false
+```
+> NOTE: pay attention to path to `make-workflows.sh`
+
+> NOTE2: pay attention to path(s) where source files are stored `files: 'PATH_REGEXP'`
--- a/make-workflows.sh/README.md
+++ b/make-workflows.sh/README.md
@@ -1,0 +1,6 @@
+---
+title: make-workflows.sh
+homepage: https://github.com/kuvaldini/make-workflows.sh
+tagline: |
+  make-workflows.sh: A tool to support YAML anchors for GitHub Actions Workflows – DRY for GHA 
+---
--- a/make-workflows.sh/install.sh
+++ b/make-workflows.sh/install.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+set -euo pipefail
+
+# function __install_make_workflows_sh() {
+#     pkg_cmd_name="make-workflows.sh"
+#     pkg_dst_cmd="$HOME/.local/bin/$pkg_cmd_name"
+#     curl -LsSf \
+#         'https://raw.githubusercontent.com/kuvaldini/make-workflows.sh/main/make-workflows.sh' \
+#         >"$pkg_dst_cmd" \
+#     && chmod +x "$pkg_dst_cmd"
+# }
+# __install_make_workflows_sh
+
+
+function __init_make_workflows_sh() {
+
+    # Every package should define these 6 variables
+    pkg_cmd_name="make-workflows.sh"
+    pkg_dst_cmd="$HOME/.local/bin/$pkg_cmd_name"
+    pkg_dst="$pkg_dst_cmd"
+    pkg_src_dir="$HOME/.local/opt/$pkg_cmd_name-v$WEBI_VERSION"
+    pkg_src_cmd="$HOME/.local/opt/$pkg_cmd_name-v$WEBI_VERSION/bin/$pkg_cmd_name"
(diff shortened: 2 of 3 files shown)


make-workflows.sh/install.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -euo pipefail
```
coolaj86, line 2: Don't set pipefail because we expect to be able to do things like test if the result of a chain with `grep` and `cut` is empty, and not error immediately.

make-workflows.sh/README.md
```diff
@@ -0,0 +1,6 @@
+---
+title: make-workflows.sh
+homepage: https://github.com/kuvaldini/make-workflows.sh
+tagline: |
+  make-workflows.sh: A tool to support YAML anchors for GitHub Actions Workflows – DRY for GHA 
+---
```
coolaj86, line 6: This should include a cheat sheet.

coolaj86:
Thanks for the submission. I took a look at your README in your main repo, but I don't have enough time right now to get a full understanding. I want to be clear on the use case before I bring this in.

I also made a few comments on the PR. :)

> It is strange to have just 168 stars.

It's because we only added the "Star on Github" two weeks ago. Before that the only reason to come to GitHub was to report a problem or add something new.

I suspect we'll hit 1k within 6 months and 10k within a year after that now that we've added the messages.

> releases.js and install.sh could be merged in one file

`releases.js` is server-side. The `install.sh` is for *nix. `install.ps1` is for Windows. Most of the heavy lifting is done in node so that the bash scripts can be relatively simple. Bash is not a great tool for looping, manipulating objects, parsing JSON, etc.

> they would look nicer just saying `webi kuvaldini/make-workflows.sh` to download and install the latest release for GH

There are 3 issues here, both of which can be overcome:

1. I don't like GitHub's politics, so I certainly don't want to make GitHub "the default". However, I have thought about this and I think the appropriate course of action would be something like this:
   ```bash
   # handle any git url, assuming 'git' to be the user by default
   webi github.com/kuvaldini/make-workflows.sh
   ```
   To make a shorthand for that I would create `ghi`, and make it possible to do
   ```bash
   curl https://webinstall.dev/ghi | bash
   # or 'webi ghi'
   ghi kuvaldini/make-workflows.sh
   ```
2. Security. We personally vet projects to some degree. We don't feature projects that we don't trust implicitly. As webi grows, it will be easy to install good tools onto your system, but it will also be easier for someone to get you to install malware.
3. Cheat Sheets. The core values of `webi` are _both_ ease-of-installation and ease-of-use.

make-workflows.sh/install.sh
kuvaldini: okay accepted.

kuvaldini:
Thank you for responses @coolaj86.

It is nice, you know what you do.

> `releases.js` is server-side

Am I right, you copy binaries to your hosting from GH, etc. using that script? Or what means server-side, i believe we do not need any intermediate server just to run a simple 10-rows nodejs script.

> Most of the heavy lifting is done in node

What you mean under "heavy lifting"? 
I see the function of releases.js is to Go to GH Releases and get link to right artifact package#platform@version. So it returns just a filtered json from api.github.com/owner/package/releases. What else?

> `webi github.com/kuvaldini/make-workflows.sh`

This looks better for me too. I've just gone the npm's way. There could be also other domains, like gitlab.com, bitbucket.com, etc.

> 3. Cheat Sheets. 

Is it possible to read them **locally**? Or it is just for reading on GH?

kuvaldini:
I have read the source script webi. I see it lacks some functionality like package removing, usual --help in console.

And the other thing. You do some work underground. Prepare installation scripts and host them on webinstall.dev. Is that process open-source, is the server open-source?

make-workflows.sh/README.md
kuvaldini: done

coolaj86:
> you copy binaries to your hosting from GH, etc. using that script

No, it selects the correct distribution based on pattern matching (which is customized for some packages) and templates that URL into the installer script.

> What you mean under "heavy lifting"?

The host target-triple and version selection is not as simple as you might think. It's easy enough to explain "it detects the os, arch, and version", but the implementation is complex.

> Is it possible to read them locally? Or it is just for reading on GH?

It's for the website. Take a look at https://webinstall.dev/rg/.

coolaj86:
> I see it lacks some functionality like package removing, usual --help in console

Yes, webi is meant to be lightweight and the scope of the project is primarily to make it easy to install things. Some of this is given in more detail on the faq page: https://webinstall.dev/faq/

> Is there dependency management?

No. The things that we host installers for don't have dependencies.

There are, however, some installers that install related tools. For example, `vim-shfmt` will install `shfmt`.

This is also talked about a bit in the FAQ.

I'm open to having a _tiny_ bit more package management - but it's gotta stay small and if lockfiles start get involved, then it's going too far.

> You do some work underground

The server is currently closed-source because that's where all the branding is, and I don't want someone "stealing" the project (I've had someone try to hijack a project before).

The bits that aren't branding are primarily url routing, templating, and caching. Some of that I'd like to move into this repo, but haven't had the time.

That said, it's the least valuable part of the project in terms of functionality and easy to reproduce from what's already here.

coolaj86:
# Preview

I've put this into beta: https://beta.webinstall.dev/make-workflows.sh

# Issues

1. I've used GitHub Actions, but I'm still not clear about what this does or how I would use it. Could you give an example of a repo that should how this is useful? It would probably be useful and appropriate to put a short demo showing inputs and outputs in the Cheat Sheet.
2. This should probably be named `gha-workflows` and install as `~/.local/bin/gha-workflows` (drop the `.sh`). `make-workflows` is a very generic name (and sounds like it could be related to `make`), but this tool is very specific to GitHub Actions
3. The cheat sheet should be run through `prettier` for proper formatting
4. The bash blocks should NOT begin with `$` (otherwise it breaks copy-and-paste to run)
5. The code blocks should be fenced with the appropriate language:
   ~~~mkdn
   ```bash
   gha-workflows --help
   ```
   ```txt
   gha-workflows:
      This script expands '*.src.yml' from $1..$[N-1] (default: REPO_ROOT/.github/)
      to $N (default:REPO_ROOT/.github/workflows/) with corresponding name '*.yml'
      Main goal is to dereference YAML anchors.
   ```
   ~~~

## Example of Demo

~~~mkdn
Say you have a folder like this:

`lsd --tree ./my-project`:
```txt
.
├── make-workflows.sh
│  ├── install.sh
│  ├── README.md
│  └── releases.js
└── .github/workflows
    ├── 1.yml
    ├── 2.yml
    └── 3.yml
```

`1.yml` may look like this:

```yml
foo:
  - bar
  - baz
  - qux
```

`2.yml` may look like this:

```yml
grualt: #foo
```

You can run gha-workflows to reference the yaml anchors:

```bash
gha-workflows .github/workflows/*.yml -o .github/workflows/main.yml
```

`main.yml`:
```yml
grualt:
  - bar
  - baz
  - qux
```
~~~

I don't actually know what this does, but I imagine it's something like that??

coolaj86:
@kuvaldini bump

kuvaldini:
@coolaj86 Hello. I was pretty busy last weeks. And I moved from macos to linux. This is still actual.

coolaj86:
@wmerfalen Is this something you can help with?

I don't understand what this is or how someone could use it. I think you've looked at GHA more than I have.

kuvaldini:
This utility is used because of lack of YAML anchor feature in GitHub actions and Workflows. It is specific to GitHub, but also could be used to unwind any YAML.
In few words anchors are named references to fields. They are used to refer (i.e take value from there) instead of coping text. This tool just genetates full unpacked YAML from yaml with anchors to make it readable by GitHub runner and workflow dispatcher.
Usebd in hyperledger/iroha and few other simpler projects.

kuvaldini:
And since last time i focused on other things and github ci is not main for me. May be it is a time to put it on a shelf. I am not sure if webi users want this tool. It is specific for a very couple of things.

coolaj86:
@kuvaldini I don't know either, but I'm willing to accept it given that two main issues are addressed:

- Can you give a precise example of its inputs and outputs so that I could understand how to use it? \
   I'd be expecting something like this: https://github.com/webinstall/webi-installers/pull/362#issuecomment-986205311
- Could you rename it to something more specific, such as `gha-make-workflow` or `gha-build-anchored-workflows`?

kuvaldini:
Okay. I will do.

wmerfalen:
> @wmerfalen Is this something you can help with?
> 
> I don't understand what this is or how someone could use it. I think you've looked at GHA more than I have.

Just seeing this. Will take a look when I can.

kuvaldini:
@coolaj86 Thank you for attention to my request. After few month this code is used I understood one important thing - it is mostly put in repository with the sources and used as an installed tool. So there is no need to forward with webinstall. 
Happy coding. Buy.

Issue #372, opened 2021-12-09
Request for eksctl for aws


xiagw:
Request for eksctl for aws

coolaj86:
`eksctl` is written in Go, so the installer should be _super_ easy to do. 

The releases are here: https://github.com/weaveworks/eksctl/releases

Anyone who wants to take this on can copy the format of https://github.com/webinstall/webi-installers/tree/main/keypairs.

Same as https://github.com/webinstall/webi-installers/issues/354.

mlopez122:
Hi! I am going to try and take on this issue and make a pull request.

mlopez122:
I have made a [pull request](https://github.com/webinstall/webi-installers/pull/374) that is ready for review.

PR #374, opened 2021-12-10
Add eksctl

mlopez122:
Following the provided instruction video I added the eksctl functionality to webinstall. I had some trouble in that I had to edit the ps1 file because when downloading the actual eksctl app, it comes as a single .exe file. 
Additionally, My native operating system is Windows, but I was using an Ubuntu install to configure the files. As a result, I kept encountering a weird occurrence where each newline was terminated with a ^M character inside of the bash script. This kept throwing errors inside of my Ubuntu client. After some troubleshooting, I found per [this stack overflow post](https://unix.stackexchange.com/questions/32001/what-is-m-and-how-do-i-get-rid-of-it) that the issue was that ^M is the carriage-return character in Windows, and it can interfere with Linux/Mac behavior. After removing the character from the bash script, the tests ran normally per the video's testing instructions.


mlopez122:
resolves #372

coolaj86:
Hey, just wanted to let you know that I appreciate this and I'll be taking a look at it within the next week or so.

I just spent a bunch of time on Webi PRs and updates a few weeks ago and I've needed to focus on some other projects for a bit to get caught up on them.

:)

eksctl/README.md
```diff
@@ -0,0 +1,51 @@
+---
+title: eksctl
+homepage: https://github.com/weaveworks/eksctl
+tagline: |
+  The official CLI for Amazon EKS
```
coolaj86, line 5: Is this actually endorsed by Amazon? I think you may mean "original" or "best", but I'm pretty sure this isn't the "official" CLI. That's written in Python and described here: https://docs.aws.amazon.com/cli/latest/reference/eks/index.html

Or am I missing something?

eksctl/README.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+To update or switch versions, run `webi example@stable` (or `@v2`, `@beta`,
```
coolaj86, line 8: Please update `example` text.

eksctl/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+```txt
+~/.config/envman/PATH.env
+~/.local/bin/eksctl
+~/.local/opt/eksctl
+```
+
+## Cheat Sheet
+
+> From the `eksctl` ReadMe:
```
coolaj86, line 24: ReadMe should be stylized README.

eksctl/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+
+To create a basic cluster:
+
+```bash
+eksctl create cluster
+```
+
+### Add Baz Highlighting
+
```
coolaj86, line 43: Replace examples sections with real examples for eksctl.

coolaj86 (changes requested):
I've requested a few changes. Please see inline comments.

Also, please run the code formatting tools before committing:

```bash
webi prettier
webi shfmt

# npx prettier@2 -w '**/*.{js,md,html}'
npm run prettier 

# shfmt -w -i 4 -sr -s ./
npm run shfmt
```

coolaj86:
This looks really good. Thanks so much for contributing.

I just have a few changes as seen above ^^. The main thing is that the docs need some fixing.

I'd love to have more cheat sheet examples, if you could. I don't use EKS, but since you do, you probably know the most common things that people want to do, and it would be great if you could add more of those. That said, it's NOT a blocking issue.

coolaj86:
P.S. This is up on beta: https://beta.webinstall.dev/eksctl

coolaj86:
bump @mlopez122

navitux:
Is this issue already solved or not ?

coolaj86:
@navitux This was left half done. The Cheat Sheet still has stuff like `example --foo --bar` in it.

Could you round that out?

Issue #350, opened 2021-11-23
[windows] Should warn if run as Admin


coolaj86:
It's important that Webi NOT be run as Admin - otherwise the ENVs get messed up and it won't work. \
(it's really late, but that should be coherent enough for me to understand in the morning)

Re #300

devPalacio:
something like this could go somewhere in the code...maybe after thanksgiving.

function Confirm-IsElevated
{
    $id = [System.Security.Principal.WindowsIdentity]::GetCurrent()
    $p = New-Object System.Security.Principal.WindowsPrincipal($id)
    if ($p.IsInRole([System.Security.Principal.WindowsBuiltInRole]::Administrator))
        { Write-Output $true }      
    else
        { Write-Output $false }   
}

if(Confirm-IsElevated)
    { throw "Please run script NOT as administrator" }

PR #383, opened 2021-12-23
fix(webi): add check for windows pwsh being run as admin

devPalacio:
#350 
I think this is a good place to put this check. Tested it and it throws a nice ugly error when you try to run the install-whatever.ps1 with powershell as admin and works in a non-elevated environment.

### DIFF
--- a/_webi/template.ps1
+++ b/_webi/template.ps1
@@ -11,7 +11,7 @@ function Confirm-IsElevated {
 }
 
 if (Confirm-IsElevated)
-{ throw "Please run script NOT as administrator" }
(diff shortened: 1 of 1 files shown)


coolaj86:
Hey, just wanted to let you know that I _really_ appreciate this!

I'll be taking a look at it within the next week or so.

I just spent a bunch of time on Webi PRs and updates a few weeks ago and I've needed to focus on some other projects for a bit to get caught up on them.

:)

coolaj86 (approved):

coolaj86:
Live!

Thank you so much! I'm so glad you figured this out. It's an absolutely wonderful contribution. Heartfelt thanks. :)

PR #398, opened 2022-02-02
Run bootstrap.ps1 from any directory

motevets:
This fixes bug whereby one could only run webi from the home directory because it created the `.local\bin` in whatever directory it was run.


coolaj86 (approved):
THANK YOU!

Sorry I didn't get to this earlier, but I'm _very_ appreciative that you found this and fixed it for us.

If you've got the Dash app, I'd be happy to send you a thank you. :)

Issue #419, opened 2022-03-08
upstream rust installer changed how PATH is exported


coolaj86:
The rust installer now needs the PATH explicitly added due to adding bash-isms to the upstream installer:

```bash
pathman add ~/.cargo/bin
```

Nasty bash-isms in the new not-env-compatible "env" file:

```bash
#!/bin/sh
# rustup shell setup
# affix colons on either side of $PATH to simplify matching
case ":${PATH}:" in
    *:"$HOME/.cargo/bin":*)
        ;;
    *)
        # Prepending path in case a system-installed rustc needs to be overridden
        export PATH="$HOME/.cargo/bin:$PATH"
        ;;
esac
```

y0rune:
Hello!
I tried to resolved that issue here (https://github.com/webinstall/webi-installers/pull/420). Please @coolaj86 check it if I understand it correct!

---
y0rune

PR #420, opened 2022-03-08
update rust installer with new PATHs

y0rune:
Hello!

I removed the default env loading into a shell config. I use for that a
`sd` and `pathman`.

If something I misunderstood please tell me what I did wrong.

Thank you,
y0rune

Re: #419

### DIFF
--- a/rustlang/install.sh
+++ b/rustlang/install.sh
@@ -2,9 +2,14 @@
 set -e
 set -u
 
+# shellcheck disable=SC2016
+
 function __install_rust() {
+
     # Straight from https://rustup.rs/
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+
+    pathman add "$HOME/.cargo/bin"
 }
 
 __install_rust


rustlang/install.sh
```diff
@@ -2,9 +2,35 @@
 set -e
 set -u
 
+# shellcheck disable=SC2016
+
 function __install_rust() {
+
+    mkdir -p "$HOME/.cargo/bin"
```
coolaj86, line 9: Move these lines 9 and 10 (mkdir and pathman) below the rust installer and that should be good enough.

We shouldn't need to edit .cargo/env.

coolaj86 (changes requested):

coolaj86:
You've still got the find and replace going, which should be removed. There's no need to edit the bash file - because fish won't use that.

rustlang/install.sh
```diff
@@ -2,9 +2,35 @@
...
     # Straight from https://rustup.rs/
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+
+    # Remove the default env file in the default shell config
+    if [ -z "$(command -v sd)" ]; then
+        webi sd
+    fi
+
+    echo ""
+    echo 'Removing from shell config '. "$HOME/.cargo/env"''
```
coolaj86, line 18: delete all of this stuff. None of this is necessary or useful.

Since `$HOME/.cargo/env` is only for bash, not for fish, it doesn't do any good to edit or remove it.

coolaj86, line 18: delete all of this stuff. None of this is necessary or useful.

Since `$HOME/.cargo/env` is only for bash, not for fish, it doesn't do any good to edit or remove it.

y0rune:
Hello!
As far I tested in the docker container with `zsh` and `fish` default shell.

- zsh
```
This path will then be added to your PATH environment variable by
modifying the profile files located at:

  /root/.profile
  /root/.bashrc
  /root/.zshenv
```

```
root@9dc724a2dce7:~# cat .zshenv
. "$HOME/.cargo/env"
```

- fish
```
This path will then be added to your PATH environment variable by
modifying the profile files located at:

  /root/.profile
  /root/.bashrc
```

So I changed the `$ZSHRC` into `$ZSHENV` and removed the fish.

coolaj86:
I don't want any changes to `.cargo/env`.

Adding the `pathman add xxxx` is sufficient and in keeping with how Webi updates PATHs.

coolaj86 (approved):

Issue #395, opened 2022-01-26
Fix simple `shellcheck` complaints (adding quotes to variables, etc)


coolaj86:
## How to see which files have warnings:

```bash
shellcheck -s sh -W 100 -e SC2034,SC2129,SC2069,SC2154,SC2164,SC2016,SC2001 */*.sh
```

#### Example output

```txt
In _webi/bootstrap.sh line 12:
    printf "Thanks for using webi to install '\e[32m${WEBI_PKG:-}\e[0m' on '\e[31m$(uname -s)/$(uname -m)\e[0m'.\n"
           ^-- SC2059 (info): Don't use variables in the printf format string. Use printf '..%s..' "$foo".


In _webi/example_install_safe_copy.sh line 6:
    if [ -n "$(command -v rsync 2> /dev/null | grep rsync)" ]; then
         ^-- SC2143 (style): Use grep -q instead of comparing output with [ -n .. ].


In arc/install.sh line 24:
        mkdir -p "$(dirname $pkg_src_cmd)"
                            ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
```

#### Example correction

```txt
In _webi/bootstrap.sh line 12:
    printf "Thanks for using webi to install '\e[32m%s\e[0m' on '\e[31m%s/%s\e[0m'.\n" \
        "${WEBI_PKG:-}" "$(uname -s)" "$(uname -m)"

In _webi/example_install_safe_copy.sh line 6:
    if command -v rsync 2> /dev/null | grep -q rsync; then

In arc/install.sh line 24:
        mkdir -p "$(dirname "${pkg_src_cmd}")"
```

## Level of Expertise Needed

Anyone with beginner bash experience could pick one (or more) of these warnings to fix across the many files in which they exist.

## Warnings that need fixing:

- [ ] SC2046 -- Quote this to prevent word splitt...
- [ ] SC2059 -- Don't use variables in the printf...
- [x] #470
- [x] ~~SC2001 -- See if you can use ${variable//se...~~
- [x] SC2005 -- Useless echo? Instead of 'echo $(...
  - [x] #460 
  - [x] #466 
- [ ] SC2143 -- Use ! grep -q instead of comparin...

saikmusunuri:
Hey, @coolaj86, I'm a first-time contributor to open source. Can you please assign me this issue?

coolaj86:
@saikmusunuri Yes.

Are you familiar with bash? Do you understand what needs to be done?

[...]

PR #421, opened 2022-03-10
fix shellcheck warnings

y0rune:
Re: #395

### DIFF
--- a/arc/install.sh
+++ b/arc/install.sh
@@ -9,19 +9,19 @@ function __init_arc() {
     ####################
 
     # Every package should define these 6 variables
-    pkg_cmd_name="arc"
+    export pkg_cmd_name="arc"
 
-    pkg_dst_cmd="$HOME/.local/bin/arc"
-    pkg_dst="$pkg_dst_cmd"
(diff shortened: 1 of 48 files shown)


y0rune:
All right I will do it ;) Sorry I added it into a issue, but it is a good idea to do it here ;)

coolaj86:
This looks good, but since it is so broad, I just need a little more time to review it.

One note: I prefer `grep` over the bash-isms because `grep` is more well-known.

I'd prefer to silence warnings about using `grep` or `sed` rather than switch to bash-isms.

y0rune:
All right! In next commits I will use grep! 
Thank you reviewing and sharing the goods way!

coolaj86:
Thanks for this. Sorry it's been getting stale.

1. Could you make the commits more atomic?
  - removing all of the `echo`s is one commit
2. We don't need exports, and we don't want bash-isms
3. Change `/bin/bash` to `/bin/sh` and remove `function` as its own commit

coolaj86:
Closing for now, waiting for new PR.

PR #426, opened 2022-03-15
docs(rustlang): Rust Hello World instructions

DBBrowne:
Please let me know if this is not correct, as that would mean that I did something wrong!

### DIFF
--- a/rustlang/README.md
+++ b/rustlang/README.md
@@ -39,7 +39,8 @@ cargo install ripgrep
 
 ```bash
 cargo new hello --bin
+pushd ./hello/
 cargo build --release
-./hello
+./target/release/hello
 > "Hello, world!"
 ```
--- a/zig/README.md
+++ b/zig/README.md
@@ -1,0 +1,213 @@
+---
+title: Zig
+homepage: https://ziglang.org/
+tagline: |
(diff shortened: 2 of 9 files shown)


coolaj86:
Thanks for the fix!

coolaj86 (approved):

rustlang/README.md
```diff
@@ -39,7 +39,8 @@ cargo install ripgrep
 
 ```bash
 cargo new hello --bin
+cd hello
```
coolaj86, line 42: If you'd like to improve this:

```bash
pushd ./hello/
```

`pushd` is what modern shells actually use when you type `cd`, and it works more reliably, even cross-platform (Windows too).

DBBrowne: Sure, clear and explicit sounds better.

`cd` works on windows, in cmd and Powershell, however it doesn't seem to add the prior directory to whatever history pushd/popd uses, so popd does not work after a cd. Interesting.

coolaj86 (approved):

coolaj86:
Thank you!

If you'd like me to buy you lunch, just drop your Đash address here and I'll send you a tip. :)

Issue #439, opened 2022-03-31
[Bug] Webi fails to install on Windows


candrewlee14:
## What were you trying to install (or what else went wrong)?

webi

## What exactly did you do?

I ran `curl.exe -A "MS" https://webinstall.dev/webi | powershell`.

## What went wrong?

I get this error:

```
PS C:\Users\candr> & "$Env:USERPROFILE\.local\bin\webi-pwsh.ps1" "webi"
curl: (35) schannel: next InitializeSecurityContext failed: Unknown error (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.

**********************************
*      IMPORTANT -- READ ME      *
*  (run the PATH command below)  *
**********************************

& : The term 'C:\Users\candr\.local\bin\pathman.exe' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At C:\Users\candr\.local\bin\webi-pwsh.ps1:71 char:7
+     & "$Env:USERPROFILE\.local\bin\pathman.exe" add ~/.local/bin
+       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\candr\.local\bin\pathman.exe:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : CommandNotFoundException
```

## Which OS did you try on?

- [ ] Mac
- [x] Windows
- [ ] Linux

## What type of computer (i.e. laptop, desktop, Raspberry Pi)?

- [x] Laptop or Desktop (amd64 / x86-64)
- [ ] Raspberry Pi (ARM-64)
- [ ] Other (please specify, if you know)

candrewlee14:
I've tried it again, and it looks like it's working. It probably was a temporary issue with Window's curl like you said. Thanks!

PR #440, opened 2022-03-31
feat(nvim): add nvim package installer

candrewlee14:
Testing on Linux worked. I'm currently unable to test on Windows due to #439. 
This will resolve this feature request: #437.

### DIFF
--- a/nvim/README.md
+++ b/nvim/README.md
@@ -1,0 +1,22 @@
+---
+title: Neovim
+homepage: https://github.com/neovim/neovim
+tagline: |
+  Neovim Vim-Fork Text Editor
+description: |
+  Neovim is a Vim fork focused on extensibility and usability
+---
+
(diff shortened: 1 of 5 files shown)


candrewlee14:
Due to both inactivity & pain points when implementing my nvim package #437, I decided to write my own cross-platform package manager called [webman](https://github.com/candrewlee14/webman), inspired by webi. It handles package installation, removal, and switching all with a single yaml file per package.

I added neovim there if you find that helpful!

coolaj86:
@candrewlee14 I happened across this today. I'll check out webman and see if I can glean any good ideas from it (I'd love to do a Webi 2.0 now that I've learned a lot more about the constraints of how the various packages need to be handled).

Non-commercial open source projects like this often benefit from a bump when you want to get something reviewed and it was missed.

If you could restore the branch that you created this from, I'd actually like to pull it in.

Issue #441, opened 2022-04-14
[Bug] vim-go: Go 1.18 drops support for `go get` (replaced by `go install`?)


coolaj86:
Originally we used `go get` because `go install` wouldn't work with go modules.

Now `go get` won't work _with_ go modules and `go install` does.

I think the solution is to replace `go get` with `go install` in https://github.com/webinstall/webi-installers/blob/main/vim-go/install.sh

# Workaround

```bash
webi vim-go

vim main.go
```

```vim
:GoInstallBinaries
```

saikmusunuri:
Hey @coolaj86, can you please assign it to me. I would like to contribute to this. Thanks!

saikmusunuri:
Hey @coolaj86, any update on this?

coolaj86:
@saikmusunuri Please do.

saikmusunuri:
Thanks! I will start working on it very soon.

saikmusunuri:
Hey @coolaj86, please review the PR.  Thanks!

saikmusunuri:
Hey @coolaj86, I once again request you to review the PR. Thanks!

candrewlee14:
Due to both inactivity & pain points when implementing my nvim package #437, I decided to write my own cross-platform package manager called [webman](https://github.com/candrewlee14/webman), inspired by webi. It handles package installation, removal, and switching all with a single yaml file per package.

coolaj86:
Closing because this was fixed and merged.

PR #442, opened 2022-04-25
[Bug] vim-go: Go 1.18 drops support for go get  #441

saikmusunuri:
Go 1.18 dropped support for `go get` so as suggested, I have replaced `go get` with `go install`. Please go through it and let me know if any change is required.


coolaj86 (approved):
Sorry for the long delay, Thanks for the help!

Drop your Đash Wallet address here and I'll send a tip / buy you lunch.

:)

PR #443, opened 2022-05-01
feat(koji): update installer and readme

its-danny:
Ref: https://github.com/its-danny/koji

### DIFF
--- a/koji/install.sh
+++ b/koji/install.sh
@@ -1,36 +1,36 @@
-#!/bin/bash
+#!/bin/sh
+# shellcheck disable=SC2034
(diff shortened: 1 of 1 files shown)


coolaj86 (approved):

coolaj86:
My apologies for taking so long on the update. Had a crazy couple of months buying a house and dealing with the economy... y'know how it goes. :)

I kept the previous cheat sheet. If there's something specific that you think ought to be different let me know, but my goal isn't to have 1:1 parity with project READMEs because webi reaches a different audience.

PR #451, opened 2022-06-04
Drop unnecessary mentions of Windows version

kidonng:
This pull request removes mentioning of Windows version. Some of them are only referring to the platform, not caring about the version, and the rest shouldn't need to either. webi packages are categorized for developers and few developers are using Windows version that old.  <sub>Remember, Windows 10 came out in July _2015_, when io.js was still around and Alphabet Inc. doesn't exist.</sub>

This also helps in case people are searching for "Windows 11", despite it being just a glorified version of Windows 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ webi_post_install       # Runs `webi_path_add $pkg_dst_bin`
 set WEBI_HOST=https://webinstall.dev
 ```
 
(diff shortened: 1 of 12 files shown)


coolaj86 (approved):
Thanks!

Drop your Đash Wallet address here and I'll send a tip / buy you lunch.

:)

PR #461, opened 2022-08-09
Added the script for creating a docker container with webi for testing

y0rune:
Hello!
I have created the PR in this repo. So I decided to create a script for testing
Webi in the Docker Container.

If you have any ideas to improve please write.

Thank you!

### DIFF
--- a/webi-docker.sh
+++ b/webi-docker.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+#shellcheck disable=SC2046
 
 VERSION_OF_UBUNTU_CONTAINER="22.04"
 WEBI_CONTAINER_FOLDER="${1:-""}"
@@ -38,7 +39,7 @@ pull_and_run_ubuntu_container() {
     docker pull jrei/systemd-ubuntu:$VERSION_OF_UBUNTU_CONTAINER
     docker run \
         -d \
-        "$([ -z "$WEBI_CONTAINER_FOLDER" ] || echo " -v ""$WEBI_CONTAINER_FOLDER"":/opt")" \
+        $([ -z "$WEBI_CONTAINER_FOLDER" ] || echo " -v ""$WEBI_CONTAINER_FOLDER"":/opt") \
         --tmpfs /tmp \
         --tmpfs /run \
         --tmpfs /run/lock \
--- a/webi-docker.sh
+++ b/webi-docker.sh
@@ -28,7 +28,9 @@ command_start() {
 }
 
 remove_old_docker_containers() {
-    container="$(docker ps --all | grep -i Webi | awk '{print $1}')"
+    container="$(docker ps --all |
+        grep -i Webi |
+        cut -d' ' -f 1)"
     if [ -n "$container" ]; then
         docker stop "$container" 2>/dev/null
         docker rm "$container" 2>/dev/null
@@ -62,7 +64,7 @@ execute_webi_in_container() {
 }
 
 ending() {
-    if [ "$WEBI_CONTAINER_FOLDER" = "" ]; then
+    if [ -z "$WEBI_CONTAINER_FOLDER" ]; then
         docker exec -it Webi mkdir -p /opt/webi
         docker exec -it Webi git clone https://github.com/webinstall/webi-installers.git /opt/webi
     else
--- a/webi-docker.sh
+++ b/webi-docker.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 VersionOfUbuntuContainer="22.04"
+WEBI_CONTAINER_FOLDER="${1:-""}"
 
 timestamp() {
     echo "[+] $(date +'%F %T') [INFO] $*"
@@ -37,7 +38,7 @@ pull_and_run_ubuntu_container() {
     docker pull jrei/systemd-ubuntu:$VersionOfUbuntuContainer
     docker run \
         -d \
-        -v "$1":/opt \
+        "$([ -z "$WEBI_CONTAINER_FOLDER" ] || echo " -v ""$WEBI_CONTAINER_FOLDER"":/opt")" \
         --tmpfs /tmp \
         --tmpfs /run \
         --tmpfs /run/lock \
@@ -60,7 +61,7 @@ execute_webi_in_container() {
 }
 
 ending() {
-    if [ -z "$1" ]; then
+    if [ "$WEBI_CONTAINER_FOLDER" = "" ]; then
         docker exec -it Webi mkdir -p /opt/webi
         docker exec -it Webi git clone https://github.com/webinstall/webi-installers.git /opt/webi
     else
@@ -74,10 +75,10 @@ ending() {
 
 main() {
     command_start remove_old_docker_containers
-    command_start pull_and_run_ubuntu_container "$@"
+    command_start pull_and_run_ubuntu_container
     command_start update_and_install_packages
     command_start execute_webi_in_container
-    ending "$@"
+    ending
 }
 
 main "$@"
--- a/webi-docker.sh
+++ b/webi-docker.sh
@@ -1,22 +1,22 @@
-#!/bin/bash
+#!/bin/sh
 
 VersionOfUbuntuContainer="22.04"
 
-function timestamp() {
+timestamp() {
     echo "[+] $(date +'%F %T') [INFO] $*"
     echo
 }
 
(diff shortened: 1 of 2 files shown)


webi-docker.sh
```diff
@@ -0,0 +1,81 @@
...
+function command_start() {
+    timestamp "Command $* has been started."
+    if ! "$@"; then
+        err "Command $* went wrong."
+        exit
+    fi
+    timestamp "Command $* has been ended."
+}
+
+function removeOldDockerContainers() {
```
coolaj86, line 28: All function declarations should follow the `snake_case()` convention.
Also remove the non-Posix `function ` since we're about to remove all of those anyway.

webi-docker.sh
```diff
@@ -0,0 +1,81 @@
...
+    docker stop "$(docker ps --all | grep -i Webi | awk '{print $1}')" >/dev/null
+    docker rm "$(docker ps --all | grep -i Webi | awk '{print $1}')" >/dev/null
+}
+
+function pullAndRunUbuntuContainer() {
+    docker pull jrei/systemd-ubuntu:$VersionOfUbuntuContainer
```
coolaj86, line 34: Why should we trust jrei? Why not use the normal Ubuntu container?

y0rune: As far I remembered I tested packaged that required the systemd in the docker container. The default ubuntu container does not have that feature.

coolaj86:
@y0rune All of the POSIX changes have been made. You should be able to test on alpine now. :)

coolaj86:
Bump @y0rune.

Will you test this now without bash?

y0rune:
All right I will test it now ;)

y0rune:
After testing the script it created in the docker container a folder `/:/opt/` when I removed from the script the mountpoint in line: https://github.com/webinstall/webi-installers/pull/461/files#diff-b8aed1a937e9424335cd5a4c1ae6e5c5c07e10f2c8ee3d491e05996881a6abeeR40.

So I will debug or resolve it ;)

y0rune:
Hello!
Firstly I will describe how the script worked:  
- Before changes:
When I run `./webi-docker.sh ` without any argument the `docker -v "":/opt/` for mounting the volume created folder `/:/opt/`
image

- After changes:
When I run `./webi-docker.sh ` without any argument the script for mounting the volume created folder `/:/opt/`
image


Secondly, I had to disable the [SC2046 – Quote this to prevent word splitting](https://www.shellcheck.net/wiki/SC2046). I tried to add in the line 42 the quotes but when I run again the script the `docker run -d ""` It did not work as should. 


Thank you!
Marcin

webi-docker.sh
```diff
@@ -0,0 +1,85 @@
...
+    docker exec -it Webi apt update
+    docker exec -it Webi apt install curl vim git npm -y
+}
+
+execute_webi_in_container() {
+    docker exec -it Webi bash -c "curl https://webinstall.dev/webi | /bin/bash"
+}
+
+ending() {
+    if [ "$WEBI_CONTAINER_FOLDER" = "" ]; then
```
coolaj86, line 65: I believe this should be `if [ -z "$WEBI_CONTAINER_FOLDER" ]`

webi-docker.sh
```diff
@@ -0,0 +1,85 @@
...
+        docker stop "$container" 2>/dev/null
+        docker rm "$container" 2>/dev/null
+    fi
+}
+
+pull_and_run_ubuntu_container() {
+    docker pull jrei/systemd-ubuntu:$VERSION_OF_UBUNTU_CONTAINER
+    docker run \
+        -d \
+        $([ -z "$WEBI_CONTAINER_FOLDER" ] || echo " -v ""$WEBI_CONTAINER_FOLDER"":/opt") \
```
coolaj86, line 42: You can use `${WEBI_CONTAINER_FOLDER:-.}` to default to `.`.

- `:-<value>` means set to `<value>` if not set, or if empty

webi-docker.sh
```diff
@@ -0,0 +1,85 @@
...
+    timestamp "Command $* has been started."
+    if ! "$@"; then
+        err "Command $* went wrong."
+        exit
+    fi
+    timestamp "Command $* has been ended."
+}
+
+remove_old_docker_containers() {
+    container="$(docker ps --all | grep -i Webi | awk '{print $1}')"
```
coolaj86, line 31: This isn't part of shellcheck, but just my own style:

- can you do this without `awk` (perhaps with `cut`)?
  - awk is a complete, standalone scripting language, like JavaScript, and I don't want to include multiple languages (perl, python, awk, js, etc) in these scripts
- split up the subshell into multiple lines
  - where possible I try to do just one thing per line so that it's easier to read hte subscript

```sh
container="$(
    docker ps --all |
        grep -i webi |
        cut -d' ' -f 1
)"
```

webi-docker.sh
y0rune: When I changed it to:
```diff
diff --git a/webi-docker.sh b/webi-docker.sh
index e56ef07..d7ba735 100755
--- a/webi-docker.sh
+++ b/webi-docker.sh
@@ -41,7 +41,7 @@ pull_and_run_ubuntu_container() {
     docker pull jrei/systemd-ubuntu:$VERSION_OF_UBUNTU_CONTAINER
     docker run \
         -d \
-        $([ -z "$WEBI_CONTAINER_FOLDER" ] || echo " -v ""$WEBI_CONTAINER_FOLDER"":/opt") \
+        -v "${WEBI_CONTAINER_FOLDER:-.}":/opt \
```
That script is not working.

```
yorune@MacBook ~/git/webi-installers webi-add-docker $ ./webi-docker.sh
[+] 2022-08-29 23:52:31 [INFO] Command remove_old_docker_containers has been started.

ea9eb78d0dc6
ea9eb78d0dc6
[+] 2022-08-29 23:52:46 [INFO] Command remove_old_docker_containers has been ended.

[+] 2022-08-29 23:52:46 [INFO] Command pull_and_run_ubuntu_container has been started.

22.04: Pulling from jrei/systemd-ubuntu
Digest: sha256:b99be1d2288bf7938e0bff3c142c8f8dc9c1cd0e6149f348217497caaf231b78
Status: Image is up to date for jrei/systemd-ubuntu:22.04
docker.io/jrei/systemd-ubuntu:22.04
docker: Error response from daemon: create .: volume name is too short, names should be at least two alphanumeric characters.
See 'docker run --help'.
[-] 2022-08-29 23:52:49 [ERROR] Command pull_and_run_ubuntu_container went wrong.

Execution time: 0h:00m:18s sec
```


I applied that echo due to the `-v ...` needs a docker command. Maybe do you know another idea to apply that feature?

coolaj86: I only use Docker when I have to, but I thought that the mapping was `<localdir>:<dockerdir>`?

Usually the docker user is `app` and the home folder is `/app` and so webi would create `/app/.local/bin/` and `/app/.local/opt/` and run the same as on desktops.

`npm` should NOT be preinstalled. If there's anything that errors out because npm isn't there, that's a bug that needs to be fixed.

Rather than `jrei/systemd-ubuntu`, the standard `Alpine` should work now.

Maybe we can get on a voice call and you can educate me on how you think this ought to work and if I can provide some feedback and maybe adjust some things in webi?

webi-docker.sh
```diff
@@ -0,0 +1,87 @@
...
+        -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
+        jrei/systemd-ubuntu:$VERSION_OF_UBUNTU_CONTAINER
+}
+
+update_and_install_packages() {
+    docker exec -it Webi echo "export TMPDIR=/compose-tmp" >>"$HOME/.bashrc"
+    docker exec -it Webi echo 'export PATH="/root/.local/bin:$PATH"' >>"$HOME/.bashrc"
+    docker exec -it Webi mkdir -p /compose-tmp
+    docker exec -it Webi apt update
+    docker exec -it Webi apt install curl vim git npm -y
```
coolaj86, line 59: `npm` should NOT be preinstalled.

coolaj86, line 59: If `curl` is not available, it should work with `wget` just fine.

y0rune:
As far I checked I can not run the alpine linux via my script so I add it into my TODO list and after that I will create another PR. Thank you @coolaj86 ;)

Issue #466, opened 2022-08-13
[Easy Refactor] Remove useless `echo`s from all scripts


coolaj86:
# What?

There are a bunch of no-op `echo`s across the scripts, which we want to remove, like this:

```diff
-    echo "$(deno --version 2> /dev/null | head -n 1 | cut -d' ' -f2)"
+    deno --version 2> /dev/null | head -n 1 | cut -d' ' -f2
```

# How

This could be done **by hand** or **by script**, and is probably as simple as installing [sd](https://webinstall.dev/sd) and committing the result of:

```sh
# remove useless echo
sd 'echo "\$\((.*)\)"' '$1' *.sh */*.sh */*/*.sh
git status
```

Reset any `typechange`-only file changes:

```
# undo 'typechange'
git status --porcelain | rg ' T' | cut -d' ' -f3 | xargs git checkout --
git status
```

Some manual checks should be done as well, for which [shellcheck](https://webinstall.dev/shellcheck) can help:

```sh
shellcheck -s sh -i SC2005 *.sh */*.sh */*/*.sh
```

# Why?

As mentioned in https://github.com/webinstall/webi-installers/issues/395, we want to fix all `shellcheck` warnings and errors.

PR #467, opened 2022-08-13
Removing a echo with command in the repo

y0rune:
Hello!
I done the issue https://github.com/webinstall/webi-installers/issues/466.
I have tested the all packages too ;) 

Please review my PR. I think it is simple :D

Thank you!


coolaj86 (approved):
LGTM

Issue #465, opened 2022-08-13
Replace all `[[ ... ]]` conditions with `[ ... ]` conditions (`shellcheck` SC3010)


coolaj86:
# What?

We need to switch out the bash-only `[[` in favor of `[` for compatibility with Alpine Docker and such. Basically something like this:

```diff
- [[ ! ... ]]
+ ! [ ... ]

- [[ ... || ... ]]
+ [ ... ] || [ ... ]

- [ ! ... ]
+ ! [ ... ] // or just `! ...` ? May need to be tested for correctness
```

It should be that only these files need updates:

```txt
_webi/bootstrap.sh
_webi/template.sh
brew/install.sh
git/install.sh
gpg-pubkey/gpg-pubkey.sh
gpg/install.sh
iterm2/install.sh
myip/myip.sh
python/install.sh
python2/install.sh
ssh-adduser/ssh-adduser.sh
test/install.sh
yq/install.sh
```

# How

This should probably be done by hand by checking each occurrence of the given warnings with [shellcheck](https://webinstall.dev/shellcheck):

```sh
# a specific file 
shellcheck -s sh -i SC3010 -- ./myip/myip.sh
```

```sh
# all files
shellcheck -s sh -i SC3010 -- *.sh */*.sh */*/*.sh
```

Each replacement should be checked against POSIX reference material, such as:
- http://mywiki.wooledge.org/BashFAQ/031
- http://mywiki.wooledge.org/BashPitfalls#A.5B.5B_.24foo_.3E_7_.5D.5D

# Why?

As mentioned in #462, we want to move away from Bash-isms and towards plain POSIX-compliant shell usage.

MoNakouzi:
@coolaj86 If this is available, can I work on it?

y0rune:
@MoNakouzi I am working on it here:https://github.com/webinstall/webi-installers/pull/468. The current status of it is blocked due to other issues. More info in the PR.

coolaj86:
Finalized with #490 and #485.

PR #468, opened 2022-08-14
Replace simple [[ ... ]] conditions with [ ... ]

y0rune:
Hello!
I used the `sd` for issue (https://github.com/webinstall/webi-installers/issues/465) 

```
# Example:
# Before: [[ cos || cos ]]
# After: [ cos ] || [ cos ]
sd '\[\[ (.*) \|\| (.*) \]\]' '[ $1 ] || [ $2 ]' *.sh */*.sh */*/*.sh

# Example:
# Before: [[ ! cos ]]
# After: ! [ cos ]
sd '\[\[ ! (.*) \]\]' '! [ $1 ]' *.sh */*.sh */*/*.sh

# Remove another cases
sd '\[\[' '[' */*.sh
sd '\]\]' ']' */*.sh
```

One line command:
```sh
sd '\[\[ (.*) \|\| (.*) \]\]' '[ $1 ] || [ $2 ]' *.sh */*.sh */*/*.sh ; sd '\[\[! (.*) \]\]' '! [ $1 ]' *.sh */*.sh */*/*.sh ; sd '\[\[' '[' */*.sh ; sd '\]\]' ']' */*.sh
```

Thank you!

### DIFF
--- a/_webi/bootstrap.sh
+++ b/_webi/bootstrap.sh
@@ -207,7 +207,7 @@ EOF
 
     chmod a+x "$HOME/.local/bin/webi"
 
-    if [ -n ${WEBI_PKG:-} ]; then
+    if [ -n "${WEBI_PKG:-}" ]; then
         "$HOME/.local/bin/webi" "${WEBI_PKG}"
     else
         echo ""
--- a/_webi/bootstrap.sh
+++ b/_webi/bootstrap.sh
@@ -182,12 +182,12 @@ __webi_main() {
         echo ""
     }
 
-    if [[ \$# -eq 0 ]] || [[ "\$1" =~ ^(-V|--version|version)$ ]]; then
+    if [ \$# -eq 0 ] || [ "\$1" =~ ^(-V|--version|version)$ ]; then
         version
         exit 0
     fi
 
-    if [[ "\$1" =~ ^(-h|--help|help)$ ]]; then
+    if [ "\$1" =~ ^(-h|--help|help)$ ]; then
         usage "\$@"
         exit 0
     fi
@@ -207,7 +207,7 @@ EOF
 
     chmod a+x "$HOME/.local/bin/webi"
 
-    if [[ -n ${WEBI_PKG:-} ]]; then
+    if [ -n ${WEBI_PKG:-} ]; then
         "$HOME/.local/bin/webi" "${WEBI_PKG}"
     else
         echo ""
--- a/_webi/bootstrap.sh
+++ b/_webi/bootstrap.sh
@@ -182,12 +182,12 @@ __webi_main() {
         echo ""
     }
 
-    if [ \$# -eq 0 ] || [ "\$1" =~ ^(-V|--version|version)$ ]; then
+    if [ \$# -eq 0 ] || [[ "\$1" =~ ^(-V|--version|version)$ ]]; then
         version
         exit 0
     fi
 
-    if [ "\$1" =~ ^(-h|--help|help)$ ]; then
+    if [[ "\$1" =~ ^(-h|--help|help)$ ]]; then
         usage "\$@"
         exit 0
     fi
--- a/_webi/template.sh
+++ b/_webi/template.sh
@@ -181,7 +181,7 @@ __bootstrap_webi() {
             # TODO wget -c --content-disposition "$my_url"
             set +e
             my_show_progress=""
-            if [ $- == *i* ]; then
+            if [[ $- == *i* ]]; then
                 my_show_progress="--show-progress"
             fi
             if ! wget -q $my_show_progress --user-agent="wget $WEBI_UA" -c "$my_url" -O "$my_dl.part"; then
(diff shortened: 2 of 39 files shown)


coolaj86:
This isn't something that can just be done automatically. It needs careful inspection of each case to make sure that it doesn't change the behavior.

POSIX `[` does not behave like Bash `[[`.

y0rune:
All right I test it all ;)

y0rune:
List of tested scripts:
- [ ] _webi/bootstrap.sh
- [ ] _webi/template.sh
- [ ] brew/install.sh
- [ ] git/install.sh
- [ ] gpg-pubkey/gpg-pubkey.sh
- [ ] gpg/install.sh
- [ ] iterm2/install.sh
- [ ] myip/myip.sh
- [ ] python/install.sh
- [ ] python2/install.sh
- [ ] ssh-adduser/ssh-adduser.sh
- [ ] yq/install.sh

coolaj86:
Thanks. I'm going to hold this one off a little longer until we're ready to switch `/bin/bash` to `/bin/sh`.

I'll try to get the others merged in today.

y0rune:
All right :+1:

coolaj86:
I found a solution to another of the `[[` bashisms:

```diff
- if [[ "$-" = *i* ]]; then
- if echo "$-" | grep -q 'i'; then
```

See <https://stackoverflow.com/questions/73412543/how-to-match-a-substring-using-posix-shell>.

_webi/bootstrap.sh
```diff
@@ -182,12 +182,12 @@ function __webi_main () {
         echo ""
     }
 
-    if [[ \$# -eq 0 ]] || [[ "\$1" =~ ^(-V|--version|version)$ ]]; then
+    if [ \$# -eq 0 ] || [ "\$1" =~ ^(-V|--version|version)$ ]; then
```
coolaj86, line 185: I don't believe that =~ is available in Posix.

Probably something like `echo "\$1" | grep -q '^(-V|--version|version)$'`

_webi/template.sh
```diff
@@ -181,7 +181,7 @@ function __bootstrap_webi() {
             # TODO wget -c --content-disposition "$my_url"
             set +e
             my_show_progress=""
-            if [[ $- == *i* ]]; then
+            if [ $- == *i* ]; then
```
coolaj86, line 184: `if echo "$-" | grep -q 'i';` should work here

brew/install.sh
```diff
@@ -7,9 +7,9 @@ function _install_brew() {
     # Straight from https://brew.sh
     #/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 
-    if [[ -n "$(uname -a | grep -i darwin)" ]]; then
+    if [ -n "$(uname -a | grep -i darwin)" ]; then
```
coolaj86, line 10: I think this should be just `if uname -a | grep -q -i darwin;`

git/install.sh
```diff
@@ -5,7 +5,7 @@ set -u
 function __init_git() {
 
     if [ -z "$(command -v git)" ]; then
-        if [[ -n "$(uname -a | grep -i darwin)" ]]; then
+        if [ -n "$(uname -a | grep -i darwin)" ]; then
```
coolaj86, line 8: same here as above

ssh-adduser/ssh-adduser.sh
```diff
@@ -56,20 +56,20 @@ function main() {
     # TODO ensure that ssh-password login is off
     my_pass="$(grep 'PasswordAuthentication yes' /etc/ssh/sshd_config)"
     my_pam=""
-    if [[ "Darwin" == "$(uname -s)" ]]; then
+    if [ "Darwin" == "$(uname -s)" ]; then
```
coolaj86, line 59: This needs to be `[ "Darwin" = "$(uname -s)" ]`

coolaj86:
`[` is actually a command, located at `/bin/[`. It is an alias of `/bin/test`. When called as `[` then the last string passed to it must be `]`.

This is the Posix standard for `test` (a.k.a. `[`): https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

Please take a look and see if you notice any non-standard uses that I missed.

y0rune:
All right I will work on it on Monday ;)

_webi/bootstrap.sh
```diff
@@ -182,12 +182,12 @@ function __webi_main () {
...
         version
         exit 0
     fi
 
-    if [[ "\$1" =~ ^(-h|--help|help)$ ]]; then
+    if [ "\$1" =~ ^(-h|--help|help)$ ]; then
```
coolaj86, line 190:
```sh
if echo "\$1" | grep -q -E '^(-h|--help|help)$'; then
```

_webi/template.sh
coolaj86: I did find that there's a "simpler" way to do this (using only Posix Shell, not grep), but it's not as easy to read:
https://stackoverflow.com/a/73419937/151312

coolaj86:
Some of the other replacements that will need to come in a follow-up...

```sh
if echo "\$1" | grep -q -E '^(-h|--help|help)$'; then
if echo "$-" | grep -q 'i'; then
if uname -a | grep -q -i darwin; then
```

coolaj86 (approved):
Needed some automatable double-quotes added, and the `[[ = > [` that need complex changes reverted

coolaj86:
@y0rune Please open a follow-up PR for the remaining changes. \
(mentioned above)

PR #483, opened 2022-08-16
fix(vps-addswap): updates root check

thebatman1:
Removes current use of EUID and UID in favour of id

Closes: #482

### DIFF
--- a/vps-addswap/install.sh
+++ b/vps-addswap/install.sh
@@ -7,7 +7,7 @@ __init_vps_addswap() {
     default_size=2G
     my_size=${1:-$default_size}
 
-    if [ "0" != "$EUID" ] && [ "0" != "$UID" ]; then
+    if [ $(id -u) -ne 0 ]; then
         echo Usage:
         echo '        sudo env PATH="$PATH" vps-addswap' $my_size
         exit 1


coolaj86:
@thebatman1 Do you know of a way to get the `id` of the login process?

I don't remember the reason I had the distinction between `UID` and `EUID`, but I think it was important for some edge case.

thebatman1:
Hi @coolaj86, yes I believe `id -ru` gives us the real user id. Let me update my PR.

I am also curious to know about the edge case you are talking about above. If you happen to remember, please share.

coolaj86:
I'm going to think through it for a minute and search through some of my old projects and blogs and see if I have a note about it.

...

(minutes later)

I think it harkens back to an inconsistency on Android when I was working on [node-installer.sh](https://git.coolaj86.com/coolaj86/node-installer.sh/src/branch/master/install.sh), which was the inspiration for Webi.

I think the issue was that `EUID` was sometimes defined when `UID` was not defined or vice-versa due to the particulars of the Android environment.

However, I believe the behavior of `id` will be more consistent across *\*nix*es, so we probably _don't_ need the "real" UID.

coolaj86:
I'll just drop that commit when I merge this in tonight / tomorrow.

coolaj86:
FYI: Never use `git merge` if you can avoid it.

`git rebase` was released very shortly after (several years ago) to fix the issues with `git merge`, but GitHub never updated their docs.

See <https://medium.com/@fredrikmorken/why-you-should-stop-using-git-rebase-5552bee4fed1>.

You use it almost exactly the same (as long as the git history hasn't become corrupted by a merge).

For example:

```sh
git rebase main
```

or, if there's a conflict you can typically just drop all duplicate commits between two branches:

```sh
git rebase -i main
```

(and then mark the duplicate / conflicting commits with `d` to drop them)

You can try that out on this PR if you'd like, or I'll just rebase it when I bring it in.

thebatman1:
Hi @coolaj86, I rebased my branch with the main branch, but push does not work since I guess no new changes were made. Please let me know if I am doing something wrong here.

Regarding the article, I am confused about the fact that you endorse git rebase, and then attach an article that says just the opposite in the very next line. I'm guessing the intent was to get me familiar with how rebase works?

coolaj86:
Oops, wrong article, I guess. The first half of it is perfect, and then he starts talking nonsense about some edge case that you'd only run into if you're forcing over main, which is a cardinal sin.

https://www.atlassian.com/git/tutorials/merging-vs-rebasing

Here's my Git Smart mini course:

https://m.youtube.com/playlist?list=PLxki0D-ilnqbXhctTFCDGH3fSN7l7ATkL

You can `git push --force` to overwrite the old commit.

coolaj86 (approved):
@thebatman1 Thanks!

:)

thebatman1:
Hi @coolaj86, here's my wallet address. Thanks for the opportunity.

XvoauTeVB2nX5jwXjrhBXECARaY7ygQd7d

coolaj86:
Sent!

Issue #500, opened 2022-08-21
Style(sh): (easy) Remove wrapping around `if` blocks (shellcheck SC2143)


coolaj86:
Re: #499 

1. Clone and enter the repo:
   ```sh
   git clone https://github.com/webinstall/webi-installers.git ./webi-installers/
   cd ./webi-installers/
   ```
2. Install [`shellcheck`](https://webinstall.dev/shellcheck) and [`shfmt`](https://webinstall.dev/shfmt):
   ```sh
   curl -sS https://webinstall.dev/shellcheck | sh
   curl -sS https://webinstall.dev/shfmt | sh
   ```
3. FOLLOW THE ON-SCREEN INSTRUCTIONS (from the installs)
4. Check all shell files for for redundant output checks
   ```sh
   shellcheck -s sh -iSC2143 *.sh */*.sh */*/*.sh
   ```
5. Update the `if` blocks according to the `schellcheck` suggestions. The necessary changes will be like this:
   ```diff
   - if [ -n "$(echo "$foo" | grep "bar")" ]; then
   + if echo "$foo" | grep -q "bar"; then
   ```
   - `grep` gets `-q`
   - `-n` will be removed
   - `-z` will be replaced with `!`
   - etc - see the instructions from `schellcheck`

rishavnandi:
Hi, I want to pick this issue up..

rishavnandi:
Hi, I want to take this issue up..
So as far as I understand I need to update the if blocks inside the shell files that are shown by the shellcheck command, right?

coolaj86:
@rishavnandi This is being worked on in #501. You could collaborate on that, or you could do a similarly easy issue, like this one: https://github.com/webinstall/webi-installers/issues/503

rishavnandi:
> @rishavnandi This is being worked on in #501. You could collaborate on that with @ShyamMahajan, or you could do a similarly easy issue, like this one: https://github.com/webinstall/webi-installers/issues/503

Thanks for the feedback I'll check it out!

Issue #474, opened 2022-08-15
Chore: Add `husky` to the repo


coolaj86:
As per <https://github.com/typicode/husky#install>:

1. Add `husky` (no transitive dependencies):
   ```sh
   npm install --save-dev husky
   ```
2. Add a new `prepare` to `package.json.scripts`:
   ```sh
   npm set-script prepare "husky install"
   npm run prepare
   ```
3. Create the new `pre-commit` script
   ```sh
   npx husky add .husky/pre-commit "npm run fmt"
   npx husky add .husky/pre-commit "npm run lint"
   git add .husky/pre-commit
   ```

TarekArar99:
Hi @coolaj86 , I'm trying to raise a PR, but when I try to commit my code it says fmt script is missing in package.json, could I get help on that ?

coolaj86:
Those commands need to be set / updated.

```sh
npm set-script fmt "npm run prettier; npm run shfmt"
npm set-script jshint "npx jshint@2.x -c ./.jshintrc */*.js"
npm set-script shellcheck "fd -t f -e sh | xargs -- shellcheck -s sh --exclude=SC2034,SC2129,SC2069,SC2154,SC2164,SC2016,SC2001 --"
npm set-script lint "npm run jshint; npm run shellcheck"
```

ShyamMahajan:
@coolaj86 I have raised a PR for this one. please check

coolaj86:
Closed via #515 and #659

PR #501, opened 2022-08-21
Fixed all shellcheck warnings and style updates #500

ShyamMahajan:
Removed wrapping around if blocks

### DIFF
--- a/_webi/example_install_safe_copy.sh
+++ b/_webi/example_install_safe_copy.sh
@@ -4,7 +4,7 @@ pkg_install() {
     (
         cd "$WEBI_TMP" 2>&1 > /dev/null
 
-        if [ -n "$(command -v rsync 2> /dev/null | grep rsync)" ]; then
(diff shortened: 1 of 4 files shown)


coolaj86:
@ShyamMahajan When you look at the output from `shellcheck` you should see that it gives different change suggestions for the different if blocks.

For example, ifs with `-z` need to be changed to `!`:

```diff
- if [ -z "$(command -v git)" ]; then
+ if ! command -v git; then
```

And only the if blocks that `shellcheck` complains about need to be adjusted. The others can be left alone.

Will you update the PR please?

ShyamMahajan:
@coolaj86 updated the PR. please review. Thanks!

coolaj86 (approved):
@ShyamMahajan Thanks!

:)

coolaj86:
@ShyamMahajan Here's another one that's pretty easy:
https://github.com/webinstall/webi-installers/issues/474

PR #510, opened 2022-08-29
docs: add new webi.sh and webi.ms

coolaj86:
We now have the domain `webi.sh` for Posix Scripts and `webi.ms` for Microsoft PowerShell.

This update various documentation to use the new domains.

This won't go live yet, but it's ready for approval.


y0rune (approved):
LGTM, but I think it should be more files due to you have done only a 9 files :D

coolaj86:
> LGTM, but I think it should be more files due to you have done only a 9 files :D

@y0rune Most of the references are generated from templates, not in the committed files.

y0rune (approved):

PR #520, opened 2022-10-02
feat: bls utility

wmerfalen:


bls/install.ps1
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env pwsh
+
+####################
+# Install bls #
+####################
+
+# Every package should define these variables
```
coolaj86, line 7: this is completely commented out

bls/releases.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var github = require('../_common/github.js');
+var owner = 'therootcompany';
```
coolaj86, line 4: this should be `dashhive`

Issue #487, opened 2022-08-18
[Doc] Simplify Caddyfile example


coolaj86:
Now that I better understand `handle`, `handle_path`, and `try_files`, we can shorten the example a little bit.

Also, we should show https on localhost.

Something closer to this:

```caddy
localhost:3443 {
    handle /api/* {
        reverse_proxy localhost:3000
    }
    handle /* {
        root * ./public/
        file_server
    }
}
```

- No `@noApi`
- No `try_files {path} {path}/ {path}/index.html`
- No `route`

caddy caddyfile config

coolaj86:
PR made: https://github.com/webinstall/webi-installers/pull/521

PR #521, opened 2022-10-04
docs(caddy): update examples, simplify config

coolaj86:
Re: https://github.com/webinstall/webi-installers/issues/487

### DIFF
--- a/caddy/README.md
+++ b/caddy/README.md
@@ -29,55 +29,112 @@ Here's the things we find most useful:
 caddy file-server --browse --listen :4040
 ```
 
+### How to serve HTTPS on localhost
+
+Caddy can be used to test with https on localhost.
+
+`Caddyfile`:
+
+```Caddyfile
+localhost {
+    handle /api/* {
+        reverse_proxy localhost:3000
+    }
+
+    handle /* {
+        root * ./public/
+        file_server
+    }
+}
+```
+
+```sh
+caddyfile run --config ./Caddyfile
+```
+
 ### How to redirect and reverse proxy
 
 Here's what a fairly basic `Caddyfile` looks like:
 
-```txt
+`Caddyfile`:
+
+```Caddyfile
 # redirect www to bare domain
 www.example.com {
     redir https://example.com{uri} permanent
 }
 
 example.com {
+    ###########
+    # Logging #
+    ###########
+
     # log to stdout, which is captured by journalctl
     log {
         output stdout
         format console
     }
 
+    ###############
+    # Compression #
+    ###############
+
     # turn on standard streaming compression
     encode gzip zstd
 
+    ####################
+    # Reverse Proxying #
+    ####################
+
     # reverse proxy /api to :3000
-    handle_path /api/* {
+    handle /api/* {
         reverse_proxy localhost:3000
     }
 
     # reverse proxy some "well known" APIs
-    reverse_proxy /.well-known/openid-configuration localhost:3000
-    reverse_proxy /.well-known/jwks.json localhost:3000
+    handle /.well-known/openid-configuration {
+        reverse_proxy localhost:3000
+    }
+    handle /.well-known/jwks.json {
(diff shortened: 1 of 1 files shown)


ryanburnette (approved):

coolaj86:
Comment. Do you have any feedback @mholt?

Preview:
https://github.com/webinstall/webi-installers/tree/docs-caddy-handle/caddy

The goal is to provide the fewest examples possible, introducing the fewest directives to learn, in the most consistent way (better to always use `handle` even with `/` than to _sometimes_ use it), with real-world-ready examples.

caddy/README.md
```diff
@@ -29,55 +29,112 @@ Here's the things we find most useful:
...
 ```
 
+### How to serve HTTPS on localhost
+
+Caddy can be used to test with https on localhost.
+
+`Caddyfile`:
+
+```Caddyfile
+https://localhost:443 {
```
mholt, line 39:
```suggestion
localhost {
```

caddy/README.md
```diff
@@ -29,55 +29,112 @@ Here's the things we find most useful:
...
+    handle /api/* {
+        reverse_proxy localhost:3000
+    }
+    
+    handle /* {
```
mholt, line 44:
```suggestion
    handle {
```

caddy/README.md
```diff
@@ -29,55 +29,112 @@ Here's the things we find most useful:
...
+        rewrite * /api/newpath{path}
+        reverse_proxy localhost:3000
+    }
+    
+    ###############
+    # File Server #
+    ###############
+    
+    # serve static files
+    handle /* {
```
mholt, line 117:
```suggestion
    handle {
```

caddy/README.md
```diff
@@ -29,55 +29,112 @@ Here's the things we find most useful:
...
+    
+    handle /* {
+        root * ./public/
+        file_server
+    }
+}
+```
+
+```sh
+caddyfile run --config ./Caddyfile
```
mholt, line 52:
```suggestion
caddy run
```

mholt (commented):
LGTM! These would be my recommendations for a simpler, less noisy config, but if you intentionally want to keep things the way they are for some reason, you are allowed to I guess.

Generally we discourage explicitly defining behavior that is already the default, since Caddy has sensible defaults, and it also makes the software unnecessarily complicated to learn and use especially when going from examples. So this is my take, i.e. I'd prefer this or do it this way if it was me writing it.

caddy/README.md
coolaj86: I've had issues with this. I believe it had something to do with browser caching / redirects.

I've found only running localhost on 443 to be more reliable than running on 80 and 443.

caddy/README.md
coolaj86: Too much mental load to have multiple ways to do the same thing.

There should be 1 obvious way to do it.

caddy/README.md
coolaj86: Same as above. Too much mental load to have multiple ways to do the same thing, and to have the most common option hidden away.

coolaj86:
@mholt I prefer the Go / Python / Zig mantras of "explicit over implicit", "Clear is better than clever", "1 obvious way to do it".

I want the lowest possible mental burden for people who are coming to webi, even if that means a few more characters to copy and paste.

That aside, are there other "every web developer is going to need to use this" type of options that should be included in the cheat sheet?

The [last one](https://webinstall.dev/caddy) was waaay to complicated, but I didn't know that there were simpler ways to do the same thing.

caddy/README.md
mholt: These two are exactly the same though. Specifying `https://` still does redirects from HTTP, and serves on port 443 by default, just as the hostname alone does.

coolaj86: Noted. I'll switch it back.

PR #547, opened 2023-01-06
added check to get user downloads directory from xdg settings

padlyuck:


_webi/template.sh
```diff
@@ -32,7 +32,15 @@ __bootstrap_webi() {
     #PKG_FORMATS=
     WEBI_UA="$(uname -a)"
     WEBI_PKG_DOWNLOAD=""
-    WEBI_PKG_PATH="${HOME}/Downloads/webi/${PKG_NAME:-error}/${WEBI_VERSION:-latest}"
+    set +e
+    WEBI_XDG_USER_DIR_BINARY="$(command -v xdg-user-dir)"
+    set -e
+    if [ -n "${WEBI_XDG_USER_DIR_BINARY}" ]; then
```
coolaj86, line 38: Thanks. A few points of discussion:

I did not know this for... until I was working on webi, but you can simplify the command check down to just:

```sh
if command -v xdg-user-dir; then
  WEBI_DOWNLOAD_DIR="$(xdg-user-dir DOWNLOAD)"
else
   WEBI_DOWNLOAD_DIR="${HOME}/Downloads"
fi
```

(and I'd like you to make that change, please :) )

But before even doing that... is it actually necessary to use `xdg-user-dir` in the first place?

What's the use case where `xdg-user-dir` works and `XDG_DOWNLOAD_DIR` does not?

Couldn't we just test for `XDG_DOWNLOAD_DIR` and have that work 99.9999% of the time? (six 9s is good enough for me)

If it's needed to account for interactive vs non-interactive shells or something, I can get behind that, but if the former works just as well in all known practical cases, I'd save the binary check until someone actually can provide the use case where it doesn't work.

I'm not as familiar with the XDG_* specs, but I assume the environment variable names are set in stone. Are the binaries as well?

Would this be a better failsafe? Or is this overcomplicating it?

```sh
if command -v xdg-user-dir; then
  WEBI_DOWNLOAD_DIR="$(xdg-user-dir DOWNLOAD)"
else
   WEBI_DOWNLOAD_DIR="${XDG_DOWNLOAD_DIR:-${HOME}/Downloads}"
fi
```

coolaj86 (changes requested):

_webi/template.sh
padlyuck: I just tested on a fresh ubuntu installation(locale ru_RU).
XDG_* variables are not exported by default, so they are not accessible from scripts.
```sh
$ cat ~/.config/user-dirs.dirs 
XDG_DESKTOP_DIR="$HOME/Рабочий стол"
XDG_DOWNLOAD_DIR="$HOME/Загрузки"
XDG_TEMPLATES_DIR="$HOME/Шаблоны"
XDG_PUBLICSHARE_DIR="$HOME/Общедоступные"
XDG_DOCUMENTS_DIR="$HOME/Документы"
XDG_MUSIC_DIR="$HOME/Музыка"
XDG_PICTURES_DIR="$HOME/Изображения"
XDG_VIDEOS_DIR="$HOME/Видео"

$ env | grep XDG
XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg
XDG_MENU_PREFIX=gnome-
XDG_SESSION_DESKTOP=ubuntu
XDG_SESSION_TYPE=x11
XDG_CURRENT_DESKTOP=ubuntu:GNOME
XDG_SESSION_CLASS=user
XDG_RUNTIME_DIR=/run/user/1000
XDG_DATA_DIRS=/usr/share/ubuntu:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
```
So I think that `xdg-user-dir` binary should be used to determine the variable value.
From this side changing to 
```sh
if command -v xdg-user-dir; then
    WEBI_DOWNLOAD_DIR="$(xdg-user-dir DOWNLOAD)"
else
    WEBI_DOWNLOAD_DIR="${HOME}/Downloads"
fi
```
looks good to me.
On another side, I'm not sure that using `WEBI_DOWNLOAD_DIR="${XDG_DOWNLOAD_DIR:-${HOME}/Downloads}"` in the `else` section is a good idea because if the `xdg-user-dir` binary is absent, but `XDG_DOWNLOAD_DIR` exists - I'm not sure that in this case system really use it.

So it looks better to me to make it as you first suggested. What do you think?

coolaj86: Perfect. All my concerns are addressed. Make the change and I'll put it on https://beta.webinstall.dev to test, and then we can bring it into `main` and https://webi.sh.

padlyuck: done

coolaj86: @padlyuck Hmm... well, I deployed to https://beta.webinstall.dev and it doesn't break on macOS, but it kinda breaks for headless Ubuntu - or at least the "Downloads" folder is instead `$HOME`. In fact, I think everything except for `DESKTOP` is `$HOME`.

I think that if `XDG_CONFIG_HOME` (or perhaps `XDG_CONFIG_DIRS`) is _NOT_ set, then even if `xdg-user-dir` exists it should be ignored.

padlyuck: I checked several cases on my server and get to the conclusion that `xdg-user-dir` may return `$HOME` path in many cases, such as:
 - `XDG_CONFIG_HOME` not set
 - `XDG_CONFIG_HOME` is set, but `$XDG_CONFIG_HOME/user-dirs.dirs` is not exists
 - `$XDG_CONFIG_HOME/user-dirs.dirs` doesn't contain `XDG_DOWNLOAD_DIR` variable or its value is empty  

Maybe we can add an additional check? something like
```sh
WEBI_DOWNLOAD_DIR="${HOME}/Downloads"
if [ -n "$(command -v xdg-user-dir)" ] && [ "$(xdg-user-dir DOWNLOAD)" != "$HOME" ]; then
  WEBI_DOWNLOAD_DIR="$(xdg-user-dir DOWNLOAD)"
fi
```

coolaj86 (approved):
@padlyuck I made a small change and retested. Thanks for the work on this!

:)

Issue #534, opened 2022-11-14
[Bug] Every few months the version update hangs


coolaj86:
# Feb 13, 2023 Update

I've deployed #565 to production.

I'm very confident (after extensive testing) that this is the final fix to intermittent failure:
- fixes incorrect cache time from 5 **seconds** (debug value) to 5 **minutes** (production value)
- always returns release info as long as usable data is available (not reset on error)

**Jan 24, 2023 Update**
 - I've deployed #551 to production.
 - I believe this refactor of release info updates is the final solution for this problem.
 - Waiting another month to close the issue...

# Original

Over the past few years we've had 2 or 3 reports like this one:
See https://github.com/webinstall/webi-installers/issues/533

Something won't install because it just hangs, and the JSON version can't be fetched.

I think that we're either missing a timeout, or mishandling it.

Come to think of it, this may be due to https://github.com/therootcompany/request.js/pull/7 and may just need to update `@root/request`.

Thankfully the hangup is per-package and not everything at once, but I'd like to get to the bottom of it.

wreszelewski:
Hello! It happens again for k9s. It looks exactly like this: https://github.com/webinstall/webi-installers/issues/533

coolaj86:
Hmmm, well the previous "fix" (https://github.com/webinstall/webi-installers/pull/535) didn't work.

I've restarted the service. I'll investigate further over Christmas break.

hwatts:
The k9s install is consistently hanging with no output at the moment

ncstate01:
I'm seeing the same thing.

coolaj86:
Restarted again just now.

I'll escalate this to high-priority. I will find a more permanent solution by this weekend at the latest.

ncstate01:
It works again.  Thank you.

coolaj86:
I've reworked the release update code.

I believe I discovered (and refactored away) a promise deadlock, and put in better failsafes for when a request hangs or fails.

It's now deployed via #551.

I'll leave this issue open for another month or two just to see if the issue arises again.

MP91:
Seems to happen again. At least I'm not able to install k9s. https://webinstall.dev/api/releases/k9s.json opens but is empty...

Don't know what causes this.

coolaj86:
Okay... I spent more time checking error conditions locally this time:
- https://github.com/webinstall/webi-installers/pull/565

Now as long as any usable data is available, the 500 will never propagate to the end user.

[...]

PR #551, opened 2023-01-24
fix: better fallbacks for release info updates

coolaj86:
Re: https://github.com/webinstall/webi-installers/issues/534

I want to get to process-level isolation, but looking over this I'm not convinced that that would do any better with just that change - something is going wrong with failed network requests.

This will do a better job at not hard failing.

There may have been a promise deadlock (those are difficult to mentally model, so I can't say for sure), but this removes the conditions that I thought were suspect.

### DIFF
--- a/_webi/transform-releases.js
+++ b/_webi/transform-releases.js
@@ -64,41 +64,55 @@ function createFormatsSorter(formats) {
 async function getCachedReleases(pkg) {
   // returns { download: '<template string>', releases: [{ version, date, os, arch, lts, channel, download}] }
 
-  function putCache() {
-    cache[pkg].promise = cache[pkg].promise.then(function () {
-      var age = Date.now() - cache[pkg].updatedAt;
-      if (age < staleAge) {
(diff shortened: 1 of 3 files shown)


_webi/transform-releases.js
```diff
@@ -64,59 +64,79 @@ function createFormatsSorter(formats) {
...
+    var age = Date.now() - cache[pkg].updatedAt;
+    if (age < staleAge) {
+      //console.debug('NOT STALE ANYMORE - updated in previous promise');
+      return cache[pkg].all;
+    }
+
+    //console.debug('DOWNLOADING NEW "%s" releases', pkg);
+    var pkgdir = path.join(installerDir, pkg);
+
+    // workaround for request timeout seeming to not work
```
coolaj86, line 88: This is the most important bit of the fix - a race between an update and returning the existing data.

This should prevent any promise deadlocks, and any failed request will just hang out in the background until it gets garbage collected.

ryanburnette (approved):

PR #555, opened 2023-01-30
hotfix: allow installers without releases

coolaj86:
In the latest update - which thankfully _didn't_ restart production when it deployed - installers with no `releases.js` all failed to install.

This fixes that.


jojobyte (approved):
Looks good

ryanburnette (commented):
Lgtm

PR #558, opened 2023-01-30
feat: add bun

coolaj86:
I started to add bun in the same way that `rustlang` was added (just use the official script)... but it's `bash` rather than `POSIX` and it doesn't have version control (yet).

So GitHub Releases for now. :)

# Preview

https://beta.webinstall.dev/bun


jojobyte (approved):
🤘🏻

PR #565, opened 2023-02-13
hotfix: another fix for github releases error responses

coolaj86:
- fixed `staleAge = 5 * 60 * 1000` (had left out the 60 for local debug last time)
- always respond when usable data is available
- light readability update around fresh/stale/expired cache times

### DIFF
--- a/_common/github.js
+++ b/_common/github.js
@@ -38,8 +38,14 @@ async function getAllReleases(
   }
 
   let resp = await request(req);
-  const gHubResp = resp.body;
-  const all = {
+  if (!resp.ok) {
+    console.error('Bad Resp Headers:', resp.headers);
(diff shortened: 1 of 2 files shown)


jojobyte (approved):
LGTM

ryanburnette:
Reviewing after the merge here, but it looks good. Our only feedback on this comes from users posting an issue if they have a failure, but we might ought to create some simulated failures to verify the user experience when those failures occur.

coolaj86:
@ryanburnette I have never been able to capture the error message (it mysteriously doesn't print, despite the `console.log()` being there), and I've never been able to reproduce it.

However I did mock error my own "boogeyman" error responses to test this with `Math.random() > 0.5` and lowering the cache time down to `5s` to retry over and over.

PR #566, opened 2023-02-18
Added the latest version of the vim plugins using the tag version

y0rune:


coolaj86:
@y0rune in order to populate the JSON versions at <https://webinstall.dev/api/releases/vim-ale.json?pretty=true> we need to have a `_common/git.js` that's similar to `_common/github-source.js`.

Would you like to try that?

coolaj86:
Closing in favor of #718, which includes what was learned here, more generically.

PR #570, opened 2023-02-23
fix(fish): ensure the config file always exists

coolaj86:
(otherwise pathman detection goes sadways)

### DIFF
--- a/fish/install.sh
+++ b/fish/install.sh
@@ -2,6 +2,14 @@
 set -e
 set -u
 
+if command -v fish > /dev/null; then
+    if [ ! -e ~/.config/fish/config.fish ]; then
+        mkdir -p ~/.config/fish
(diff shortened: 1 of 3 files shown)


fish/install.sh
```diff
@@ -2,6 +2,14 @@
 set -e
 set -u
 
+if command -v fish > /dev/null; then
+    if [ ! -e ~/.config/fish/config.fish ]; then
```
jojobyte, line 6: Why not wrap this in a function if its gonna call the exact same code twice?

jojobyte (approved):
See my question, but otherwise approved.

fish/install.sh
coolaj86: My general rule is that if something is relatively small, I'll write it twice. If I write it 3 times, then I'll consider an abstraction.

PR #571, opened 2023-02-23
doc(bat) add Files section

HacDan:
Add files section for `bat` to README.md

### DIFF
--- a/bat/README.md
+++ b/bat/README.md
@@ -7,6 +7,13 @@ tagline: |
 
 To update or switch versions, run `webi bat@stable` (or `@v0.18`, `@beta`, etc).
 
+### Files
+```text
+~/.config/envman/PATH.env
+~/.config/bat/config
+~/.local/opt/bat/
+```
+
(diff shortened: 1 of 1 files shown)


coolaj86:
Also needs `~/.config/bat/config`, if you don't mind.

HacDan:
I was referencing the `install.sh` script when writing up the Files section but I may have missed the creation of the file.

coolaj86:
It doesn't actually create it yet, but it's one of those files that you'd want to see at-a-glance to know that it exists at all.

It _should_ create it and the fact that it doesn't doesn't make it less valuable to know about it.
You pickin' up what I'm putting down?

HacDan:
I'm smelling what you're steppin' in, don't you worry. 

I pushed a Fixup commit. Should be all set. If you'd rather I squash them, I suppose I can do that.

coolaj86 (approved):
Approved. Rebased and squashed

PR #573, opened 2023-02-24
doc(rg) add Files section

HacDan:
Added Files section for ripgrep. 

I added the referenced config file from ripgrep's [docs](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file). I don't believe Webi is creating it and honestly, I didn't even know you could have a config file for ripgrep.


coolaj86 (approved):
I used this as an excuse to squeeze in some doc updates that came to mind as well.

PR #582, opened 2023-03-06
fix: remove outdated config on formatters, and detect by file type

coolaj86:
It looks like the desired behavior now exists upstream, so we can remove our special edits.

~/.vim/pack/plugins/start/vim-shfmt/plugin/shfmt.vim

```vim
" ...
		autocmd BufWritePre *.sh Shfmt
		autocmd FileType sh autocmd BufWritePre <buffer> Shfmt
" ...
```

### DIFF
--- a/vim-prettier/prettier.vim
+++ b/vim-prettier/prettier.vim
@@ -3,8 +3,15 @@
 " from webinstall.dev/vim-prettier "
 """"""""""""""""""""""""""""""""""""
 
-" format as-you-type is quite annoying, so we turn it off
-let g:prettier#autoformat = 0
+" Change Log
+"
+" 2023-03-06:
+"   - run when filetype matches javascript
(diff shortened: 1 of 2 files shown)


coolaj86:
I found this helpful to put into my `~/.vimrc` for having `vim-ale` detect `bun`, `deno` and `zx` as javascript. It also seems to cause `vim-prettier` to detect `deno` and `zx` as javascript, but not `bun` 🤷‍♂️.

```vim
" set filetype to javascript by shebang
au BufRead,BufNewFile * if getline(1) =~ '#!/.*\<\(bun\|deno\|zx\)\>' | setfiletype javascript | endif
```

jojobyte (approved):
👍🏻

coolaj86:
Tracking at https://github.com/prettier/vim-prettier/issues/340

PR #594, opened 2023-04-05
Docs-update

alexravenna:
I fixed an error in the `fd` readme (most likely copy-paste) and updated the `nerdfont` readme to correspond to the current version of Windows Terminal.


coolaj86 (approved):
@alexravenna Thanks for these fixes!

Drop a Đash Wallet address here (or via DM) and I'll send a tip / buy you lunch.

:)

Issue #597, opened 2023-04-30
[Bug]


marcpmichel:
## What were you trying to install (or what else went wrong)?

webi on Windows 11

## What exactly did you do?

curl.exe https://webi.ms/arc | powershell

## What went wrong?

error downloading pathman.exe

```
curl "https://rootprojects.org/pathman/dist/windows/amd64/pathman.exe"
curl: (7) Failed to connect to rootprojects.org port 443 after 2931 ms: Couldn't connect to server
```

the url is from :[install.ps1]( https://github.com/webinstall/webi-installers/blob/main/pathman/install.ps1)

scott-morris:
It appears that all of https://rootprojects.org is denying connections. May need to find another source for this executable.

coolaj86:
Looking into this now. That should have been git.rootprojects.org.

coolaj86:
Fixed. @scott-morris Try now.

A few years ago the Linux installer was updated to use the git release URL, but the Windows installer was not. Despite that, the old release location remained up for all these years, until now.

My apologies to @marcpmichel and others who were having trouble with this earlier and I didn't see the report.

Issue #428, opened 2022-03-17
[Idea] Update PATH on Windows natively


coolaj86:
`pathman` was created in part because it actually wasn't possible to update a PATH on Windows in `cmd.exe`.

Turns out, they've added a way to do that in Powershell:
https://ziglang.org/learn/getting-started/#setting-up-path-on-windows

```pwsh
[Environment]::SetEnvironmentVariable(
   "Path",
   [Environment]::GetEnvironmentVariable("Path", "User") + ";C:\your-path\zig-windows-x86_64-your-version",
   "User"
)
```

coolaj86:
Closed because this is rolled up into ^^ that issue.

PR #599, opened 2023-05-09
ref!: replace pathman with PowerShell [Environment]::SetEnvironmentVariable

coolaj86:
At long last, https://github.com/webinstall/webi-installers/issues/428 has come to fruition (thanks in large part to #597 and #598).

> When it is broken, it is the right time to fix it. - Chinese Fortune Cookie I once had

### DIFF
--- a/_webi/template.ps1
+++ b/_webi/template.ps1
@@ -42,27 +42,87 @@ New-Item -Path .local\opt -ItemType Directory -Force | out-null
 # {{ baseurl }}
 # {{ version }}
 
-function webi_add_path
+function webi_path_add($pathname)
 {
-    Write-Host ''
-    Write-Host '*****************************' -ForegroundColor red -BackgroundColor white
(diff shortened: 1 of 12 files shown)


jojobyte (approved):
LGTM

coolaj86:
I did have one concern: How _recent_ is this API to PowerShell?

Will current users on older versions of Windows be negatively impacted?

The earliest references I can find are 2014:
- https://archive.org/details/promicrosofthdin0000sark/page/66/mode/2up?q=PowerShell+SetEnvironmentVariable
- https://archive.org/details/twoscoopsofdjang0000gree_d1l4?q=PowerShell+SetEnvironmentVariable

It looks like it was available in .Net (but not PowerShell?) at least as early as 2010:
- https://archive.org/details/professionaliron0000muel/page/216/mode/2up?q=PowerShell+SetEnvironmentVariable

I assume that this means that all Windows 10 users who are the types of users that would be using Webi (i.e. they've run Windows Update sometime within the last 10 years - not banks) should have this available.

Damn the torpedoes, we're taking it live!

Issue #606, opened 2023-05-18
[Bug] nerdfont install location has moved


coolaj86:
Was:
- https://github.com/ryanoasis/nerd-fonts/raw/master/patched-fonts/DroidSansMono/complete/Droid%20Sans%20Mono%20Nerd%20Font%20Complete.otf

Now is:
- https://github.com/ryanoasis/nerd-fonts/raw/master/patched-fonts/DroidSansMono/DroidSansMNerdFontMono-Regular.otf
- https://github.com/ryanoasis/nerd-fonts/releases/download/v3.0.1/DroidSansMono.tar.xz

Older version:
- https://github.com/ryanoasis/nerd-fonts/releases/tag/v2.3.3

## Solution

Swap `master` for `v2.3.3`:

```diff
- https://github.com/ryanoasis/nerd-fonts/raw/master/patched-fonts/DroidSansMono/complete/Droid%20Sans%20Mono%20Nerd%20Font%20Complete.otf
+ https://github.com/ryanoasis/nerd-fonts/raw/v2.3.3/patched-fonts/DroidSansMono/complete/Droid%20Sans%20Mono%20Nerd%20Font%20Complete.otf
```

PR #607, opened 2023-05-18
fix(nerdfont): pin version to tag 'v2.3.3' rather than 'master' (now v3.x)

coolaj86:
workaround / fix for #606: use the old (working) version

(a better fix would be to switch to using the releases, but overwriting the install function to use the Fonts folder)


jojobyte (approved):
Looks good.

PR #615, opened 2023-05-28
vim-airline installer

ryanburnette:
- [x] vim-airline installer
- [x] vim-airline-themes installer

### DIFF
--- a/vim-airline-themes/README.md
+++ b/vim-airline-themes/README.md
@@ -2,19 +2,33 @@
 title: vim-airline-themes
 homepage: https://github.com/vim-airline/vim-airline-themes
 tagline: |
-  The official theme repository for vim-airline
+  vim-airline-themes are the 1st-party themes for vim-airline
 ---
+
+To update (replacing the current version) run `webi vim-airline-themes`.
+
+### Files
+
+These are the files / directories that are created and/or modified with this
+install:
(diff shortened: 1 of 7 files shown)


ryanburnette:
I know the cheatsheet is sparse, but vim-airline is a plugin that modifies the interface and I personally use defaults. The screenshot is all that I need in the cheat sheet.

ryanburnette:
The `airline.vim` config file is actually empty if you just install vim-airline from webi, but I like to create it in case the user wants to use it, and later if you install vim-airline-themes it'll add a line to that file.

ryanburnette:
@coolaj86 Do you mind taking a look at the themes installer with me? My intent here is to:

- place the airline.vim file if it is not present
- if it is present, look to see if a theme is set, if one is set, leave it alone, if one isn't set, append the theme setting line from the theme installer airline.vim file

I believe I have it working but it might be good for me to add some echoes.

Unless you have some edits regarding that, this is ready to merge.

ryanburnette:
I know you'll also want more elaborate cheat sheets, but I use this basically with defaults. You install it and get the new behavior. I don't have much else to add.

coolaj86 (approved):
Muchas garcias!

(I just made a few style and format edits)

PR #620, opened 2023-06-09
feat: add sshd-prohibit-password

coolaj86:
moving the prohibit password checks from ssh-adduser into their own utility

To-Do:
- [ ] `ssh-utils` needs `chmod a+x sshd-*`

### DIFF
--- a/ssh-adduser/install.sh
+++ b/ssh-adduser/install.sh
@@ -3,18 +3,26 @@ set -e
 set -u
 
 __install_ssh_adduser() {
-    my_cmd="ssh-adduser"
-
-    rm -f "$HOME/.local/bin/${my_cmd}"
-
+    # ssh-adduser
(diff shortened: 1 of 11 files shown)


ryanburnette:
Would be cool to have something like `ssh-adduser-hard` to preserve the previous behavior

coolaj86:
> Would be cool to have something like `ssh-adduser-hard` to preserve the previous behavior

The behavior changes for the single file as part of `ssh-utils`, but remains the same for `curl https://webi.sh/ssh-adduser`:

https://github.com/webinstall/webi-installers/pull/620/files#diff-122e5f4d928d49151910bf3266b53c779883ad1a082fd7a3419e6f1e3c934511R21-R25

I'm thinking of something like `vps-init` for the future as a standalone command.

ryanburnette (approved):

Issue #272, opened 2021-07-23
Wanted: ssh-authorize


coolaj86:
# Step 1

Create and publish a cross-platform command to add a key to `~/.ssh/authorized_keys`, and set the permissions properly if they are incorrect.

```sh

USAGE

    ssh-authorize [string-or-file-or-url] [comment]

EXAMPLES

    ssh-authorize single-key.pub

    ssh-authorize https://coolaj86.com/authorized_keys

```

Bonus: check that the keys aren't already added.

This would be a great project to write in Rust - especially if it doesn't require the extra visual studio runtime libraries - or otherwise Go.

I suppose separate commands in Powershell and Bash could suffice too, if need be.

# Step 2

Make a PR to add the installer to Webi, using https://github.com/webinstall/webi-installers/tree/main/keypairs or https://github.com/webinstall/webi-installers/tree/main/ssh-pubkey as a template.

ryanburnette:
Is this helpful? https://man.openbsd.org/ssh-keyscan.1

ryanburnette:
Something like ...

```bash
ssh-authorize https://ryanburnette.com/pubkey
```

... would be cool.

coolaj86:
> Is this helpful? https://man.openbsd.org/ssh-keyscan.1

No. That gets the keys of the servers you want to connect to, not the ones you want to connect from.

coolaj86:
@ryanburnette How's that 1Password thing work again?

PR #621, opened 2023-06-10
feat: add ssh-authorize

coolaj86:
As discussed in #272 

I've run various tests with good and bad files, urls, and existing and non-existing folder structure.

It's good. It's ready.

### DIFF
--- a/ssh-authorize/README.md
+++ b/ssh-authorize/README.md
@@ -1,0 +1,61 @@
+---
+title: SSH Authorize
+homepage: https://webinstall.dev/ssh-authorize
+tagline: |
+  Add to your SSH Authorized Keys from a string, file, or url.
+---
+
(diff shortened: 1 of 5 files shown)


ryanburnette:
I humbly submit that ssh-authorize scripts should have its own repo.

I don't think it's best to have source code that is the functionality of the script in our installers repo.

Issues the functionality also don't belong on the installers repo issues.

ryanburnette (approved):

ryanburnette (approved):

coolaj86:
> I humbly submit that scripts like ssh-adduser, ssh-authorize, and so on should have their own repos.

Agreed. I was thinking about this. I'd like to create an ssh-utils repo and move this all out at some point. Just a matter of priorities.

PR #627, opened 2023-06-21
fix(dashd): turn on indexes

coolaj86:
Many of the most important RPCs can't work without indexing turned on.

This will likely increase the system requirements as well.

### DIFF
--- a/dashcore-utils/dash.example.conf
+++ b/dashcore-utils/dash.example.conf
@@ -1,3 +1,8 @@
+txindex=1
+addressindex=1
+timestampindex=1
+spentindex=1
+
 [main]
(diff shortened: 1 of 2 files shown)


jojobyte (approved):

coolaj86:
Note this makes it very, _very_, ***very*** slow.

- `test` only went from 1h to 2.5 hours
- `main` went from 4h to... 32h(?) - currently sitting at 8h/25%

However, nothing is maxing out - CPU, Bandwidth, RAM, and Storage all have plenty of headroom. It's very unclear what's causing this.

coolaj86:
Also: `rpcconnect` should be set to the same value as `rpcbind`. Redundant, but required.

PR #633, opened 2023-07-07
ref(dashcore-utils): add `zmqpubhashchainlock` to dash.example.conf

coolaj86:
1. as a workaround for inaccurate results from the `getaddressutxos` RPC it will be necessary to have `zmqpubhashchainlock` enabled so that `zmqpubrawtxlock` can be used to get the missing utxos and know when to expel them from memory.
2. update README with QuickStart


jojobyte (approved):
Looks good

PR #634, opened 2023-07-15
doc(ffmpeg): Add DNS resolution note and example

jojobyte:
Add note about loss of DNS resolution due to statically linked `glibc`  and example scenario


coolaj86 (approved):

coolaj86:
Will you squash this with the message in the format of `fix(ffmpeg): ...`

PR #641, opened 2023-07-30
ref(vim-essentials): use vim-ale's built-in prettier support

coolaj86:
So it turns out we haven't needed `vim-prettier` since `vim-ale`.

We just needed to update `~/.vim/plugins/ale.vim` with `let g:ale_fix_on_save = 1`.

### DIFF
--- a/beyond-shell/README.md
+++ b/beyond-shell/README.md
@@ -59,7 +59,6 @@ recommended.
   - [vim-smartcase](/vim-smartcase)
   - [vim-spell](/vim-spell)
   - [vim-ale](/vim-ale)
-  - [vim-prettier](/vim-prettier)
   - [vim-shfmt](/vim-shfmt)
   - [vim-whitespace](/vim-whitespace)
 - Commandline Tools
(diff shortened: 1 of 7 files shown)


coolaj86:
Needs some other updates:

```diff
- let g:ale_echo_msg_format = '[%linter%] %s [%severity%]'
+ let g:ale_echo_msg_format = '%code%: %s [%linter%] [%severity%]'
```

```diff
+ let g:ale_fixers = {
+ \  'markdown': ['prettier'],
+ \}
```

coolaj86:
Since (as far as I know) we don't have an easy way to modify the config to support different fixers, I'm actually going to let this one go.

`vim-prettier` does a better 0-config, out-of-the-box job than `vim-ale` on this one and less difficult-to-change config is better.

Note: perhaps we should move lists like that to their own file each:

```text
./vim/pulgins/ale.vim
./vim/pulgins/ale-linters.vim
./vim/pulgins/ale-fixers.vim
```

That would make it easier to ad-hoc parse and generate the list.

```text
aletool get fixers
aletool add fixers.markdown prettier
```

PR #649, opened 2023-09-11
doc(node): add serviceman examples

coolaj86:
# Preview at

https://github.com/webinstall/webi-installers/tree/coolaj86-node-serviceman/node


jojobyte (approved):
Looks good.

PR #655, opened 2023-10-09
Add MSVC Runtime (VC Redist) (vcruntime140.dll)

coolaj86:
Many of the Windows programs are compiled with MSVC, which requires the 25mb of Windows that Microsoft just won't install for you.

This makes that simpler.

### DIFF
--- a/msvc-runtime
+++ b/msvc-runtime
@@ -1,0 +1,1 @@
+vcruntime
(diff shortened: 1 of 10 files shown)


jojobyte (approved):
Looks good.

y0rune (approved):
LGTM!

PR #656, opened 2023-10-11
fix(vim-ale): show linter and code in a way that works for jshint and shellcheck

coolaj86:
Very simple output change

```text
[jshint] : 'await' is not defined. [E]
```

```text
[shellcheck] SC2154: my_var is referenced but not assigned. [W]
```


jojobyte (approved):
👌🏻

Issue #645, opened 2023-08-28
[Bug] k9s on ppc64 gets ppc64le binary


camberkenpas:
## What were you trying to install (or what else went wrong)?
Trying to install k9s on ppc64le.

## What exactly did you do?
I simply ran:
` curl -sS https://webinstall.dev/k9s | bash`

## What went wrong?
The installer thinks the platform is "ppc64" when in fact it is "ppc64le": 
```
$ curl -sS https://webinstall.dev/k9s | bash

Thanks for using webi to install 'k9s@stable' on 'Linux/5.15.0-41-generic/ppc64le'.
Have a problem? Experience a bug? Please let us know:
        https://github.com/webinstall/webi-installers/issues

Lovin' it? Say thanks with a Star on GitHub:
        https://github.com/webinstall/webi-installers

Error: no 'k9s' release for 'linux' on 'ppc64' as one of 'tar,exe,xz,git,dmg' by the tag 'stable'
       'k9s' is available for 'macos,linux,windows' on 'amd64,arm64,ppc64le,s390x' as one of 'tar,zip'
       (check that the package name and version are correct)

       Double check at https://webinstall.dev/api/releases/k9s@stable.tab
$ uname -m
ppc64le

```
Probably the installer script is checking for the architecture slightly incorrectly.

## Which OS did you try on?

- [ ] Mac
- [ ] Windows
- [x] Linux

## What type of computer (i.e. laptop, desktop, Raspberry Pi)?

- [ ] Laptop or Desktop (amd64 / x86-64)
- [ ] Raspberry Pi (ARM-64)
- [x] ppc64le

coolaj86:
Yes, detecting these variations can be tricky and I've never actually tested ppc because I don't have any ppc devices.

I could probably make ppcle I higher priority than ppc.

coolaj86:
Hey @camberkenpas, thanks for the report. This is now fixed by #657

camberkenpas:
I have confirmed that this is working, thank you!

PR #657, opened 2023-10-11
fix(ppc64le): sort ppc64le BEFORE ppc64

coolaj86:
RE: #645


ryanburnette (approved):

y0rune:
@coolaj86 BTW I think you can also remove the branch ;)

PR #662, opened 2023-10-12
Add ffuf and feroxbuster (security based tooling)

copyleftdev:
added a security tool ffuf feroxbuster, it is commonly used by security researchers, I figured having an varied set of tools for webi would get more users in hope that translates to [Stars]

### DIFF
--- a/ffuf/README.md
+++ b/ffuf/README.md
@@ -1,0 +1,60 @@
+---
+title: ffuf
+homepage: https://github.com/ffuf/ffuf
+tagline: |
+  Fuzz Faster U Fool: A fast web fuzzer written in Go.
+---
+
+To update or switch versions, run `webi ffuf@stable` (or `@v2`, `@beta`, etc).
+
+### Files
+
+These are the files / directories that are created and/or modified with this
+install:
+
+```text
+~/.config/envman/PATH.env
+~/.local/bin/ffuf
+```
+
+## Cheat Sheet
+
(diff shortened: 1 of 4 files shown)


coolaj86:
@copyleftdev I rebased these on master, so you'll need to `git pull --rebase` for further edits.

Can you update these examples to use something like pastebin or some companion tool that makes it more clear what they're doing? Or a honey pot site or local honey pot tool?

Also, these are missing ` ```sh ` code blocks for the examples.

copyleftdev:
@coolaj86 I updated the ffuf readme

coolaj86 (approved):
I squashed (`git rebase -i main`) all the commits and updated the Cheat Sheet after doing some research on ffuf myself.

coolaj86:
@copyleftdev Will you open a new PR with `feroxbuster` and model the cheat sheet after my updates to `ffuf`?

https://webinstall.dev/ffuf

We want the Cheat Sheet to be something approachable for newbies that aren't already familiar with similar tools.

Things that were challenging in learning about fuff:
- it's not clear what a wordlist is, what format it's in, or where to get one
- it's not clear that `FUZZ` is a user-selectable substitution variable
- it's not clear what the results are expected to be

The installers should be pretty much copy/paste, but the Cheat Sheet is where it really takes someone who is familiar with the tool, and make it accessible to those who aren't.

PR #665, opened 2023-10-12
added xsv installer

copyleftdev:
added xsv installer

### DIFF
--- a/xsv/README.md
+++ b/xsv/README.md
@@ -5,7 +5,7 @@ tagline: |
   xsv: A fast CSV command line toolkit written in Rust.
 ---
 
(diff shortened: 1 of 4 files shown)


coolaj86 (approved):
I've been interested in `xsv` for a long while and this gave me the inspiration to do a deep dive and really learn how to use it. Thanks for that.

Also, that is to say that the way I've modified the Cheat Sheet on this one is NOT typical of what I expect, but I think `xsv` is one of the special cases that really deserves more info.

I think the author's README is actually pretty good, so a lot of this is redundant... but it's what I did. :)

Also: fixed Windows installer.

Issue #251, opened 2021-06-09
Update docs for dotenv-linter v3.0


coolaj86:
dotenv-linter v3.1 is out. A number of options have changed. Our cheatsheet should be updated according to the new usage.

## Our Cheat Sheet

Deployed: https://webinstall.dev/dotenv-linter/

Source: https://github.com/webinstall/webi-installers/blob/main/dotenv-linter/README.md

## Top-Level Usage

```bash
dotenv-linter --help
```

```txt
dotenv-linter 3.1.0
Mikhail Grachev <work@mgrachev.com>
Lightning-fast linter for .env files

USAGE:
    dotenv-linter [FLAGS] [OPTIONS] <input>... [SUBCOMMAND]

FLAGS:
    -h, --help         Prints help information
        --no-color     Turns off the colored output
    -q, --quiet        Doesn't display additional information
    -r, --recursive    Recursively searches and checks .env files
    -v, --version      Prints version information

OPTIONS:
    -e, --exclude <FILE_NAME>...    Excludes files from check
    -s, --skip <CHECK_NAME>...      Skips checks

ARGS:
    <input>...    files or paths [default: /Users/me]

SUBCOMMANDS:
    compare    Compares if files have the same keys [aliases: c]
    fix        Automatically fixes warnings [aliases: f]
    list       Shows list of available checks [aliases: l]
```

### Fix Usage

```bash
dotenv-linter fix --help
```

```txt
dotenv-linter-fix
Automatically fixes warnings

USAGE:
    dotenv-linter fix [FLAGS] [OPTIONS] <input>...

FLAGS:
    -h, --help         Prints help information
        --no-backup    Prevents backing up .env files
        --no-color     Turns off the colored output
    -q, --quiet        Doesn't display additional information
    -r, --recursive    Recursively searches and checks .env files

OPTIONS:
    -e, --exclude <FILE_NAME>...    Excludes files from check
    -s, --skip <CHECK_NAME>...      Skips checks

ARGS:
    <input>...    files or paths [default: /Users/me]
```

### Compare Usage

```bash
dotenv-linter compare --help
```

```txt
dotenv-linter-compare
Compares if files have the same keys

USAGE:
    dotenv-linter compare <files>...

FLAGS:
    -h, --help        Prints help information
        --no-color    Turns off the colored output
    -q, --quiet       Doesn't display additional information

ARGS:
    <input>...    Files to compare
```

## Changes from Documentation

```bash
dotenv-linter fix
```

[...]

PR #676, opened 2023-10-17
docs(dotenv-linter): Update Cheat Sheet to be compatable with dotenv-linter v3.1

thegreatestgiant:
Updated the Cheat Sheet for dotenv linter as according to #251 
I also changed --skip to shorthand -s which is less verbose but I kept compare verbose (I didn't do dotenv-lint c .env1 .env2)


coolaj86:
@thegreatestgiant What happened here?

thegreatestgiant:
I created a pull request
then I created another without thinking that would apply to this one
and then my attempted solution closed it

PR #680, opened 2023-10-17
updated fuff readme

copyleftdev:
added more details to readme cheat sheet


coolaj86:
@copyleftdev what I was asking for is a cheat sheet for feroxbuster.

I'll incorporate some of this into the ffuf cheat sheet, but my changes here were to serve as an example for docs for the other packages you've added.

PR #682, opened 2023-10-18
docs(dotenv-linter): Update Cheat Sheet to be compatable with dotenv-linter v3.1

thegreatestgiant:
Updated the Cheat Sheet for dotenv linter as according to https://github.com/webinstall/webi-installers/issues/251
I also changed --skip to shorthand -s which is less verbose but I kept compare verbose (I didn't do dotenv-lint c .env1 .env2)


coolaj86 (approved):
Thanks!

PR #706, opened 2023-10-25
ref!: move webi bootstrap to own file

coolaj86:
The `webi` script has been embedded in the bootstrap script from the beginning.

Since embedded scripts can't be linted or formatted, it's gotta get pulled out.

That also mean that the bootstrap script needs to include the ability to download, which means we might as well rewrite it.

- remove embedded Webi script (it's already in `webi/webi.sh` now)
- disambiguate the names of the bootstrap and template scripts (`curl-pipe-bootstrap.tpl.sh`)
- rewrite bootstrap script, with colors!

# notes to self

~~I learned a little `sed` trick that makes it possible to embed a file within another file.~~

~~So now, rather than having about a hundred `\"` and `\$` and `\\` sprinkled through a precarious string template, `webi` itself can actually pass linting!~~

Except that trick doesn't work when streamed from the network...


coolaj86:
Except that trick doesn't work when streamed from the network...

coolaj86:
Okay, this isn't going to happen right now.

The same chicken-and-egg problem that applied in the old scenario still applies presently.

If only there was some way to get `$(uname -srm)` from curl at the get-go without a literal `-H "User-Agent: $(uname -srm)" \` ...

coolaj86:
Okay, this actually does need to happen soon. There's more than one way to skin a cat. Re-opening.

HacDan (commented):
Isn't this in the other PR? #738 

### EDIT:

Alright, I'm confused, I'm seeing undoing in two places. What am I missing?

coolaj86:
@HacDan This one was based on that one. Now that that one is pulled in, the commit list in this one has been updated. Should make more sense now.

HacDan (approved):
LGTM. Thank you for clarifying the supporting commit.

PR #713, opened 2023-10-27
Fix: Add if condition to welcome message

### LINKED ISSUE #711: Too much love (only show the support message once)
```text
Thanks for using webi to install 'node@20.8' on 'Darwin/arm64'.
Have a problem? Experience a bug? Please let us know:
        https://github.com/webinstall/webi-installers/issues

Lovin' it? Say thanks with a Star on GitHub:
        https://github.com/webinstall/webi-installers
```

Needs to be silenced in meta packages (`vim-essential`) and packages that install multiple things (`prettier`, `redis-commander`, `xcaddy`).

We could probably add a "export WEBI_IS_LOVED=true" at the top of a meta package and not show the message for any of the child installers.

## Ideal, Per Install

For a meta package it will still be very helpful to know where things went wrong.

And so... we might want to book-end multi-installers with the _Starting_ and _Completed_ messages.

```text
Fetching 'vim-essentials[@ver]' for 'Darwin/arm64'...
```

```text
Installing 'vim-essentials[@ver]' on 'Darwin/arm64'
```

## Ideal, Per Meta Install

Ideally we should detect when there is any sort of exit 1 and immediately display the error message with the debug info and url.

Then at the end it should print the message.

However, in reality we probably want
- print "Have a problem?" _before_ any possible error \
  (in case something is catastrophically wrong)
- print "Lovin' it?" _after_ success

```
Thanks for using webi to install 'vim-essentials' on 'Darwin/arm64'.
Have a problem? Experience a bug? Please let us know:
        https://github.com/webinstall/webi-installers/issues
```

```
Lovin' it? Say thanks with a Star on GitHub:
        https://github.com/webinstall/webi-installers
```

thegreatestgiant:
This should fix #711 
It adds an if condition of
```sh
test -z "$WEBI_WELCOME"
```
to the install script and should only run if WEBI_WELCOME doesn't have a value
Should I add another (AND or OR) condition to make sure it `=true`?

P.S. - Also I'm having a hard time testing my code, I tried `WEBI_PKG=vim-essentials bash _webi/bootstrap.sh` but it only displays the love message once. Any recommendations?

### DIFF
--- a/_webi/bootstrap.sh
+++ b/_webi/bootstrap.sh
@@ -13,15 +13,17 @@ __install_webi() {
     #WEBI_PKG=
     #WEBI_HOST=https://webinstall.dev
     export WEBI_HOST
-
-    echo ""
-    printf "Thanks for using webi to install '\e[32m%s\e[0m' on '\e[33m%s/%s\e[0m'.\n" "${WEBI_PKG-}" "$(uname -s)/$(uname -r)" "$(uname -m)"
-    echo "Have a problem? Experience a bug? Please let us know:"
-    printf "        \e[2m\e[36mhttps://github.com/webinstall/webi-installers/issues\e[0m\n"
-    echo ""
-    printf "\e[35mLovin'\e[0m it? Say thanks with a \e[1m\e[33mStar on GitHub\e[0m:\n"
-    printf "        \e[36mhttps://github.com/webinstall/webi-installers\e[0m\n"
-    echo ""
+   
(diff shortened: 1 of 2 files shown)


coolaj86 (approved):
Hmm... looks like I had done the work for this and just forgot to close the loop (er... condition) on that one.

Thanks!

coolaj86:
That wasn't quite the whole fix. I had to add the WEBI_WELCOME=true and export as well - but then it fixed it. :)

PR #718, opened 2023-10-29
feat: use `git` as source for releases

coolaj86:
# Preview

- https://next.webinstall.dev/vim-commentary
- https://next.webinstall.dev/api/releases/vim-commentary.json?pretty=true

Also:

```sh
# update webi
curl https://webi.sh/ | sh

# view vim-commentary info on 'next'
export WEBI_HOST='https://next.webinstall.dev'
webi --info vim-commentary
```

```sh
# install via 'next'
curl https://next.webinstall.dev/vim-commentary | sh
```

# Details

There are a number of types of utilities that only need a git tag:

- VimScript (`vim-ale`)
- POSIX Shell Scripts (`aliasman`, `duckdns.sh`)
- PowerShell

This allows us to use `git` as a source of release info.

```sh
node ./_common/git.js
```

```json5
{
  name: 'ale-v3.3.0',
  version: 'v3.3.0',
  date: '2022-12-25T20:30:21.000Z',
  ext: 'git',
  command: 'git clone --depth=1 --single-branch --branch v3.3.0 https://github.com/dense-analysis/ale.git'
}
```

Re: https://github.com/webinstall/webi-roadmap/issues/17

PR #720, opened 2023-10-30
style+lint: simplify `test`s to use exit code & file size rather than output/contents

coolaj86:
Cleaning up a few things I've found along the way while preparing #718.

- `[ -n "$(command -v foo)"]` => `command -v foo > /dev/null`
- `[ -n "$(cat ./foo.txt)" ]` => `test -s ./foo.txt`
- `echo $(foo) | cut -d' ' -f2` => `foo | tr -s ' ' | cut -d' ' -f2`

The number and complexity of these changes were right in that middling number where I decided not to automate it, but I have triple checked each line of change to make sue the `-z` vs `-n` and `!` were all transitioned correctly.

### DIFF
--- a/sudo/install.sh
+++ b/sudo/install.sh
@@ -4,14 +4,20 @@ set -u
 
 __init_sudo() {
 
-    if [ -z "$(command -v sudo)" ]; then
-        echo >&2 "Error: on Linux and BSD you should install sudo via the native package manager"
-        echo >&2 "       for example: apt install -y sudo"
-        exit 1
-    else
+    if command -v sudo > /dev/null; then
         echo "'sudo' already installed"
+        exit 0
     fi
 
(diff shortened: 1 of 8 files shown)


HacDan (approved):
LGTM! Happy path first =)

sudo/install.sh
```diff
@@ -4,14 +4,14 @@ set -u
 
 __init_sudo() {
 
-    if [ -z "$(command -v sudo)" ]; then
-        echo >&2 "Error: on Linux and BSD you should install sudo via the native package manager"
-        echo >&2 "       for example: apt install -y sudo"
-        exit 1
-    else
+    if command -v sudo > /dev/null; then
         echo "'sudo' already installed"
```
coolaj86, line 8: I missed a spot: needs `exit 0`

PR #729, opened 2023-11-05
fix(windows): use home directory, not current directory

coolaj86:
This one snuck by for a long time.

I mean, it _usually_ worked.


HacDan (approved):
I'm curious how this worked for so long... thanks for putting this together! 

Gotta be pure coincidence. 

1. Open CMD
2. run `webi go`
3. ...
4. Close CMD

Anyway, LGTM. I did test on Win11 just to make sure.

PR #730, opened 2023-11-05
feat: add TinyGo

coolaj86:
I finally rebuilt `pathman` in a way that came out to a reasonable size.

And if I use a tool... it goes on Webi.


HacDan (approved):
README looks good!

Did not test the install script, but logic is sound.

PR #732, opened 2023-11-05
fix(busybox): remove extraneous regex escape

coolaj86:
Those backslashes... they'll getcha!

This only affected error messages on BusyBox sed

XKCD 1638

(and the backslash for the escape you thought you needed, but didn't, of course)


HacDan (approved):
LGTM!

PR #735, opened 2023-11-06
ref(api): use await to de-indent some async functions

coolaj86:
This is just a whitespace change. No functional differences.


HacDan (approved):
Looks good. Thank you for putting this together!

PR #736, opened 2023-11-06
refactor `curl-pipe-bootstrap.tpl.sh` and `install-package.tpl.sh`

coolaj86:
`webi`, the command, has grown too big to be embedded any longer.

Plus, once we move it out, we may be able to drastically simplify the installer template as well. We shall see.

### DIFF
--- a/_webi/curl-pipe-bootstrap.tpl.sh
+++ b/_webi/curl-pipe-bootstrap.tpl.sh
@@ -222,8 +222,10 @@ fn_sub_home() { (
 ); }
 
 main() { (
-    fn_show_welcome
-    export WEBI_WELCOME=true
+    if test -z "${WEBI_WELCOME:-}"; then
+        fn_show_welcome
+    fi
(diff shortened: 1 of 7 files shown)


HacDan (approved):

coolaj86:
I ended up adding a bugfix needed for some of the suggestions in https://github.com/webinstall/webi-installers/pull/768 and reordering some of the functions in comments in `curl-pipe-bootstrap.sh` so that the changes to `install-package.tpl.sh` would match up nicely.

It's a lot of commits but not actually much changed:
https://github.com/webinstall/webi-installers/compare/8aa665b...e4c4fd8?expand=1

In the end, do the reordering and name cleanup in the bootstrap first allowed me to copy nearly the whole file into the package install and get just this as the diff:
- https://github.com/webinstall/webi-installers/pull/736/commits/b56229ca62fa557e189bcd11e3282a83f4acdadb

PR #737, opened 2023-11-07
various prep cleanup

coolaj86:
RegExp replace changes have been tested as follows:

```sh
curl -sS 'https://webinstall.dev/api/installers/node@20.sh?formats=tar,zip,xz,git,dmg,pkg&os=darwin&arch=arm64&libc=libc' > install-node-prod.sh

curl -sS 'https://next.webinstall.dev/api/installers/node@20.sh?formats=tar,zip,xz,git,dmg,pkg&os=darwin&arch=arm64&libc=libc' > install-node-next.sh

diff install-node-*.sh
```

I manually compared, fixed, and retested. The templated fields are working as they were before, and as expected.


HacDan (approved):

coolaj86:
I also added a few things that have to be deployed in lockstep in upcoming PRs:
- webi script in its own installer
- bootstrap renderer brought into webi core (it was dangling in the webserver)

PR #738, opened 2023-11-07
fix: don't switch template file name yet

coolaj86:
Oops! A commit with the new filenames was pulled in by mistake!

This undoes that, which shouldn't come until #736 

Live in `hotfix`.


HacDan (approved):
LGTM

Issue #698, opened 2023-10-23
add bash, zsh, fish completions to `webi`


Lockszmith-GH:
Adding bash-competion to the webi cli tool would be a great addition and quality-of life improvement.

I understand it's probably a non-trivial impelmentation, but having a package name auto-complete would go along way to promote the installation of the cli tool.

coolaj86:
How about just a `webi list` or `Webi search`?

I'm all for command completion, but:
1. I've never done it
2. That would require a regularly updated local cache

I'd happily accept a PRs that generate Bash, ZSH, and Fish completions.

Could you do any of that?

coolaj86:
Oooh! I happened across how `pyenv` does completions whilst working on #702, and I LOVE IT!

- command generates its on completions on-the-fly
- can fetch from the web while generating
- works across all shells at once, give that the outputs are correct

## Examples

- `~/.pyenv/completions/pyenv.bash`
- `~/.pyenv/completions/pyenv.fish`
- `~/.pyenv/completions/pyenv.zsh`

### `~/.pyenv/completions/pyenv.bash`:

```sh
function __fish_pyenv_needs_command
  set cmd (commandline -opc)
  if [ (count $cmd) -eq 1 -a $cmd[1] = 'pyenv' ]
    return 0
  end
  return 1
end

function __fish_pyenv_using_command
  set cmd (commandline -opc)
  if [ (count $cmd) -gt 1 ]
    if [ $argv[1] = $cmd[2] ]
      return 0
    end
  end
  return 1
end

complete -f -c pyenv -n '__fish_pyenv_needs_command' -a '(pyenv commands)'
for cmd in (pyenv commands)
  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a \
    "(pyenv completions (commandline -opc)[2..-1])"
end
aj@biggoron ~> cat ~/.pyenv/completions/pyenv.bash
_pyenv() {
  COMPREPLY=()
  local word="${COMP_WORDS[COMP_CWORD]}"

  if [ "$COMP_CWORD" -eq 1 ]; then
    COMPREPLY=( $(compgen -W "$(pyenv commands)" -- "$word") )
  else
    local words=("${COMP_WORDS[@]}")
    unset words[0]
    unset words[$COMP_CWORD]
    local completions=$(pyenv completions "${words[@]}")
    COMPREPLY=( $(compgen -W "$completions" -- "$word") )
  fi
}

complete -F _pyenv pyenv
```

[...]

PR #742, opened 2023-11-09
feat(webi): add shell completions (#698)

RubenRam:
Re: #698

@coolaj86 I've added an "init" option in a similar fashion to "pyenv init" or similar programs that populates the local config of each supported shell with completion functions in order to query the available packages.

However, I've noticed that querying every time to the server (webi --list) can be a bit laggy and from a user perspective might not be great. I suggest maybe redirecting the query to a local cache/file with the list of available packages. Let me know what you think when you try it.

Last thing, for debugging purposes, I've commented the --list option warning as it was overwriting the completions options print and lead to confusion. 

# Todos (AJ)

- [x] generate something like `list.txt` in place of `sitemap.xml`
- [x] show `--list`, `--help`, etc?
- [ ] ~~show aliases `powershell (pwsh)`, `iterm (iterm2)`, etc~~

# Two-Stage Diff (AJ)

This should have been two PRs (one prep, one the actual work), but instead we can look at these two diffs:

- Cleanup Refactor: https://github.com/webinstall/webi-installers/compare/7eec48e...f10be58?expand=1
- Important Diff: https://github.com/webinstall/webi-installers/compare/f10be58...948030b?expand=1


coolaj86:
It may be a few days before I check this.

Here's my quick thoughts on a caching strategy:

```
webi_get_list() { (
    mkdir -p ~/.local/share/webi/var/
    if ! test -r ~/.local/share/webi/var/last_update; then
        echo '0' > ~/.local/share/webi/var/last_update
    fi

    my_stale_age=600
    my_now="$(date -u '+%s')"
    my_then="$(cat ~/.local/share/webi/var/last_update)"
    my_diff=$((my_now - my_then))

    if test "${my_diff}" -gt "${my_stale_age}"; then
        webi --list > ~/.local/share/webi/var/cache.part
        mv ~/.local/share/webi/var/cache.part ~/.local/share/webi/var/cache
        echo "${my_now}" > ~/.local/share/webi/var/last_update
    fi

    cat ~/.local/share/webi/var/cache
); }
```

coolaj86:
This is pretty sweet!

Screenshot 2023-11-12 at 1 53 52 AM

coolaj86:
@RubenRam Will you figure out how we can show annotations? For example: I want to show that `iterm` is an alias for `iterm2` like:

```text
iterm (iterm2)
iterm2
powershell (pwsh)
pwsh
rg
ripgrep (rg)
```

RubenRam:
@coolaj86, sorry for getting back to you just now.

Well thought of the async update, I did some local tests which caching but that definitely did the trick for a smooth experience.

I see as well that you've modified accordingly the webi list function, great. Do you think it would would be of any use adding an option to the script (--no-cache) to call directly the **_webi_list_uncached** function? I could picture a situation were checking the latest content for dev/debug purposes would be helpful. Let me know your thoughts on this.

As for the alias feature you mention, I'm not entirely sure what you are asking me. Would you like the alias to show up during the completions or just another print? The static list part got me confused. However, trying to bring ideas here, I think it would make sense adding an alias field to sitemap.xml to later process them in the **webi_list** function.

Lastly, glad you like the feature!

coolaj86 (approved):
I've tested extensively on macOS and Linux across bash, zsh, and fish.

Refactored headliner-fashion into smaller functions with early returns.

Thanks for getting the ball rolling on this @RubenRam!

PR #743, opened 2023-11-10
3. feat(ssh-authorize): powershell edition for Windows

coolaj86:
So. Much. Work.

But it works! On Windows!

## How install `ssh-authorize` on Windows

Once this PR is pulled in:

```pwsh
curl.exe https://webi.ms/ssh-authorize | powershell
```

## How to Set Permissions and add SSH Public Keys for Windows

Just running the command with no arguments will correctly set the permissions:

```sh
ssh-authorize
```

## How to Quickly Add Authorized SSH Keys on Windows

Like this:

```sh
ssh-authorize https://github.com/johndoe.keys
```

Or this:

```sh
ssh-authorize ./johndoe.pub
```

It adds them to both of
- `$HOME\.ssh\authorized_keys`
- `${Env:ProgramData}\ssh\administrators_authorized_keys`

And uses `icacls.exe` to properly shift the permissions around so that `sshd` will run without complaining (and without disabling authorized keys).

This DOES NOT require `Repair-AuthorizedKeyPermission` from `OpenSSHUtils`, but if you'd like to see how it's done we have our own `Repair-UserAuthorizedKeyPermission` and `Repair-AdminAuthorizedKeyPermission`:

https://github.com/webinstall/webi-installers/pull/743/files#diff-2049d699094d380ddb32e50b937ce7f239980b4070fec2df8ef4c046f1fa5697R19-R86

### DIFF
--- a/_webi/install-package.tpl.ps1
+++ b/_webi/install-package.tpl.ps1
@@ -1,8 +1,6 @@
 #!/usr/bin/env pwsh
 #350 check if windows user run as admin
 
-$OriginalPath = $Env:Path
-
 # this allows us to call ps1 files, which allows us to have spaces in filenames
 # ('powershell "$Env:USERPROFILE\test.ps1" foo' will fail if it has a space in
 # the path but '& "$Env:USERPROFILE\test.ps1" foo' will work even with a space)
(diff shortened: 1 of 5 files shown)


HacDan (commented):
I'll give this a test tonight. Looks good though.

HacDan (approved):

PR #753, opened 2023-11-12
ref(pwsh): swap alias since `pwsh` is the command name, not `powershell`

coolaj86:
The name of the command (not the project) should be the canonical reference.

- delete `pwsh` alias
- rename `powershell` to `pwsh`
- create `powershell` alias to `pwsh`


HacDan (approved):
LGTM. Thank you for putting this together. I appreciate the added ARM changes =)

PR #759, opened 2023-11-14
Fix flutter release URLs

coolaj86:
Found the release URLs doing a devtools inspect on the releases site:

https://docs.flutter.dev/release/archive?tab=windows

Thankfully that worked... for now.


coolaj86:
Also: https://github.com/flutter/flutter/issues/138422#issuecomment-1811045358

HacDan (approved):
LGTM. Good catch on a happinstance! 

I had a PR half done for this and then caught up on the stream and noticed you decided to just push it through, so thank you!

PR #763, opened 2023-11-14
fix(ssh-adduser): check if SSH_ADDUSER_AUTO is set

coolaj86:
Needs `${SSH_ADDUSER_AUTO:-}` when used standalone.


HacDan (approved):
LGTM. I had to look up the `:-` as I was unsure of what it did.

coolaj86:
See also: https://github.com/mvdan/sh/issues/970

PR #764, opened 2023-11-15
Only call load.sh once, even when .bashrc sources .profile

Lockszmith-GH:
Changed load.sh so that it will load only once.

On fresh debian and ubuntu systems, I have .profile calling .bashrc.

I like it that webi installs the load.sh source-calling in both, but I'd like it to load only once.

Added a 'singleton flag' of sorts to prevent load.sh from loading more than once in a session.

### DIFF
--- a/_webi/package-install.tpl.sh
+++ b/_webi/package-install.tpl.sh
@@ -285,24 +285,24 @@ __bootstrap_webi() {
         fi
 
         if ! test -e ~/.config/envman/load.sh; then
-            # shellcheck disable=SC2016
-            {
-                echo 'if [ -z "$ENVMAN_LOADED" ]; then'
-                echo '    ENVMAN_LOADED=1'
+            cat << LOADENVMAN > ~/.config/envman/load.sh
+if [ -z "\$ENVMAN_LOADED" ]; then
+    ENVMAN_LOADED=1
 
-                echo '# Generated for envman. Do not edit.'
-                echo 'for x in ~/.config/envman/*.env; do'
-                echo '    my_basename="$(basename "${x}")"'
-                echo '    if [ "*.env" = "${my_basename}" ]; then'
-                echo '        continue'
-                echo '    fi'
-                echo ''
-                echo '    # shellcheck source=/dev/null'
-                echo '    . "${x}"'
-                echo 'done'
+    # Generated for envman. Do not edit.
+    for x in ~/.config/envman/*.env; do
+        my_basename="\$(basename "\${x}")"
+        if [ "*.env" = "\${my_basename}" ]; then
+            continue
+        fi
 
-                echo 'fi'
-            } > ~/.config/envman/load.sh
+        # shellcheck source=/dev/null
+        . "\${x}"
+    done
+
+fi
+LOADENVMAN
+
         fi
 
         if command -v sh > /dev/null; then
--- a/_webi/package-install.tpl.sh
+++ b/_webi/package-install.tpl.sh
@@ -827,10 +827,12 @@ fn_envman_init() { (
         mkdir -p ~/.config/envman/
     fi
 
+    # Note: the variables $BASH, $ZSH_NAME, etc are always empty
+    # because the active shell is always sh when this script runs
     fn_envman_init_load_sh
-    fn_envman_init_shell sh .profile
-    fn_envman_init_shell bash .bashrc
-    fn_envman_init_shell zsh .zshrc
+    fn_envman_init_shell sh '.profile' '.ash_history'
+    fn_envman_init_shell bash '.bashrc' '.bash_history'
+    fn_envman_init_shell zsh '.zshrc' '.zsh_sessions'
 
     if command -v fish > /dev/null; then
         fn_envman_init_load_fish
@@ -871,15 +873,21 @@ LOAD_SH
 fn_envman_init_shell() { (
     a_shell="${1}"
     a_rc="${2}"
+    a_history="${3:-_history_file_doesnt_exist}"
+    a_login_shell="$(basename "${SHELL:-}")"
 
     if ! command -v "${a_shell}" > /dev/null; then
         return 0
     fi
 
-    if ! test -e ~/"${a_rc}"; then
-        return 0
+    # .bashrc and .zshrc no longer exist by default on macOS
+    if ! test -e ~/"${a_rc}" && ! test -e ~/"${a_history}"; then
+        if test "${a_login_shell}" != "${a_shell}"; then
+            return 0
+        fi
     fi
 
+    touch -a ~/"${a_rc}"
     if grep -q -F '/.config/envman/load.sh' ~/"${a_rc}"; then
         return 0
     fi
--- a/_webi/package-install.tpl.sh
+++ b/_webi/package-install.tpl.sh
@@ -258,6 +258,8 @@ __bootstrap_webi() {
         )"
 
         my_export="export PATH=\"$my_path_export:\$PATH\""
+
+        touch -a ~/.config/envman/PATH.env
         if grep -q -F "${my_export}" ~/.config/envman/PATH.env; then
             return 0
         fi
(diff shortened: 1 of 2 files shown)


coolaj86:
This is a good solution. Thank you.

I'm in the middle of a big round of tedious testing of target host matching, but once I'm at a good pause point I'll test this and pull it in.

Notes to self:
- we don't need an export because the variable is being sourced rather calling
- may or may not escape within the string to keep the alignment

coolaj86:
@Lockszmith-GH Will you check the "allow maintainers to make edits box?" for this fork I want to do make some changes:

- rename variable and value to `ENVMAN_LOAD='loaded'`
- early return rather than wrapping
- re-align the spaces
- possibly other refactoring while I'm in this general area of code

Lockszmith-GH:
> @Lockszmith-GH Will you check the "allow maintainers to make edits box?" for this fork I want to do make some changes:

Didn't find a checkbox, but I invited you as contributor (Write permissions).
Let me know if that isn't enough.

coolaj86:
Thanks. I've rebased it on `main`.

I'm testing this out and it's not as simple as it would seem.

I think that your approach does solve the specific problem of .bashrc sourcing .profile without causing any unwanted side-effects.

However, I've been playing around trying to find a solution that also works when I'm in `fish` and I open up `bash` or `sh`, or I'm in `bash` and open up `zsh`.

At present, I think there are only going to be 4 files:
- `alias.env` (each line is something like `alias ll='ls -lAhF'`, can't be exported)
- `function.env` (ex: `ll() {( ls -lAhF );}`)
- `ENV.env` (ex: `export VISUAL='vi'`)
- `PATH.env` (ex: `PATH="$HOME/bin:$PATH"`)

So I may need two variables `g_envman_load` and `ENVMAN_LOAD`, and to unroll the loop into 3 distinct cases, or handle `alias.env` as a special case. Or treat *UPPER*s as exports and *lower*s as non-export.

Lockszmith-GH:
how about the following pattern, This should workd on sh, bash and zsh (I hanv't jumped on the fish bandwagon yet, so I'll leave that to you):

Instead of:
```bash
if [ -z "\$ENVMAN_LOADED" ]; then
    ENVMAN_LOADED=1
```

would this work?
```bash
if ! (printf '%s' "$ENVMAN_LOADED" | grep -q ":?${SHELL}\$"); then
    ENVMAN_LOADED=:"${SHELL}"
```

I had a similar 'dynamicly loaded dotfiles dir' scheme, but my library was personal, so never had to deal with the intricacies of dealing with multiple different environments to this extent.

I reaload my shell profile with the following, so I always 'lose' the environment - it's the only solution I found for clearing aliases
```bash
_r() { exec $SHELL -l; }
```

coolaj86:
@Lockszmith-GH Check out the latest commit I pushed.

I kinda rewrote the whole thing, but I think it's more readable than it was before, and solves the problem from every direction.

I've done some light testing and I'm gonna do a bit more right now.

coolaj86 (approved):
Thanks @Lockszmith-GH for your work on this. I'm pulling it in, along with my refactor that solves the problem for the other shells.

If you create and drop a Đash Wallet address here (or DM me) I'd love to send you a tip/buy you lunch as thanks for getting this rolling.

:)

coolaj86:
P.S. Here's how I decided to solve it generally:

```sh
# Generated for envman. Do not edit.

# always make sure the files exist so that we can grep without error
touch -a ~/.config/envman/PATH.env
touch -a ~/.config/envman/ENV.env
touch -a ~/.config/envman/alias.env
touch -a ~/.config/envman/function.env

# handle the global exports that persist between processes
# (such as calling a bash or sh script from a zsh or bash parent)

# ENV first because we may use it in PATH
test -z "${ENVMAN_LOAD:-}" && . ~/.config/envman/ENV.env
test -z "${ENVMAN_LOAD:-}" && . ~/.config/envman/PATH.env
export ENVMAN_LOAD='loaded'

# Handle interactive shell variables (aliases, functions) separately
# (because they must be loaded for each process - they don't inherit from the parent)

# function first because we may use it in alias
test -z "${g_envman_load_sh:-}" && . ~/.config/envman/function.env
test -z "${g_envman_load_sh:-}" && . ~/.config/envman/alias.env
g_envman_load_sh='loaded'
```

Lockszmith-GH:
> If you create and drop a Đash Wallet address here (or DM me) I'd love to send you a tip/buy you lunch as thanks for getting this rolling.

All good, really appreciate the offer, but conisder my contribution support of the project.

Lockszmith-GH:
I liked the fact that it would load `*.env`, it allowed me to dynamically add files there, instead of editing shell scripts.

Are you still open for changes here?
Would a `*.env` vs `*.ENV` work?

coolaj86:
What about just doing a

We could do something like `session.d` or `env.d`

coolaj86:
We could do something like `session.d/` or `env.d/`, but one problem I see with that, in the general case, is that you actually don't know what order things will load in.

What about just adding one-liners to `ENV.env`?

What's your use case?

I've been working on `envman` as a stand-alone tool because the goal was never for Webi to manage the Environment, but I needed something that would work reliably for adding the path to installs.

I honestly didn't think anyone would actually start using the envman directory, I was just putting it in place for automation and tooling.

Lockszmith-GH:
Correct, but that can be resolved with `sort` and proper naming.

`00_path.env` will make sure it runs before `another_env_file.env`

We're dealing with a shell, but you can set the rules.
`sort` seems to be a safe approach without overcomplicating things.

I currently have a scheme of my own, and if your vision for envman is different, I'll stick with mine - but for a moment there, I saw somebody doing something similar, and "I got excited" 🤓 

What I do is this:
I have a sub-dir, like your .config/envman, where I iterate over all of the .sh files there (this allows me to switch something off, by just prefixing it with soimething else, like `.off`)
My load script loads the relevant environment/functions/aliases based on the tooling.

This makes managing the environment with a tool like `chezmoi` very easy, instead of managing blocks of code in monolithic file, I have snippets I can turn on/off - also great for troubleshooting.

I actually went one step deeper, I have an additional set of dirs, `<shell>.pre` and `<shell>.post`, so I get to initialize each shell properly in case unique solutions are required per shell, but that can be overkill.

so something like this:
```
~/.config/envman:
+--- 00_init.env
+--- 11_PATH.env
+--- reload_func.env
+--- 99_bash_completions.env
+--- 99_zsh_completions.env
```

This way you get to control the order of invocation.
if you want to split the re-load, then you can use differnt extensions or split to different sub-dirs, like you suggested.

But going back here, that's my vision of tooling I already have in place, if you end up implementing somehting like this, I might shift to your implementation, just to consolidate/minimize on my end, but you stated a different vision.

coolaj86:
I'm just now realizing that I made a mistake with this.

`function.env` should have been sourced as `function.sh` and `function.fish`.

And the `session.d` would need to be `session.zsh.d`, etc.

What is your workflow like that you get benefit from disabling a shell file?

I get the logical completeness of all of what you're saying, but what's the day-to-day (or project-to-project) utility?

Where the cost is low and the benefit is high, I'm all for including features, but the complexity of this stuff is so deceptive. I think "I'll make a 2 minute change" and then after 3 hours of testing and debugging...

And I haven't looked into nushell yet, but particularly because it works on Windows I want to try it out. I doubt one generic tool could be an effective shell manager, but I'm hoping a single tool can be a good ENV manager.

Lockszmith-GH:
My end goal was to develop something like webi for myself. I never got there. (That's also why I'm really excited about the possibilites of your tool).

The dir structure would allow me to just add the supporting 'startup files' into this dir structure on-demand:
When a tool is installed, supporting files are added, when removed, supporting files are out.
Restarting the shell (with that `_r` function) and I'm ready to go.

Shell completion, history management, prompt customizations, personal environment, environment secrets/tokens etc.

Separate files also allows me to share a dotfiles environment with a team, while everyone can have their local modification without affecting others, becuase every thing is in separate files.
[This is how](https://github.com/twpayne/chezmoi/discussions/2574) I manage the shared/personal-envrionments.

The on/off is mostly a troubleshooting technique, useful in-development to just take something out quickly.

Lockszmith-GH:
My env is currently a mess, as I'm in trnasition between 'worlds' - worked in the linux world for the past couple of years, and now I'm back in 'windows' world, so things have shifted. When I get it stabalized - it will finally be functional in both Windows and Linux properly, I'll be happy to share it.

PR #756, opened 2023-11-14
100% accurate classification of target host triples (i.e. `linux-unknown-aarch64-musl`)

coolaj86:
This is a "Trust me bro" PR. There are too many changes to test them all exhaustively, but I've tested as exhaustively as I can - especially with the `lint-xyz` files - and it's been in prod where 1 bug and 1 backwards-compat issue have been discovered.

Any other issues will just have to be fixed.

# Summary 🚀

<https://next.webinstall.dev/> can reliably detect, classify, sort, and match *all* known user agent strings, build asset names, and version identifiers.

- [x] 100.0% version classification (as per #783)
- [x] 100.0% host target (User-Agent / uname) identification ( #778 )
- [x] 100.0% build target (package) identification
- [x] OpenBSD, FreeBSD, Dragonfly, etc support ( #780 )

# Tasks

In tandem with https://github.com/webinstall/webi-build-classifier.

- [x] Sort builds by triplet
- [x] User-Agent to canonical triplet
- [x] Canonical triplet to acceptable triplets
- [x] Return matches, sorted by version and extension
- [x] cold-boot cache (don't factor in stale time) 
- [x] warm lazy refresh (always return existing data, update in bg)
- [x] hourly random refreshes 
- [ ] https://github.com/webinstall/webi-installers/issues/609

# ARM Notes

> the h stands for hard float. the l stands for little endian. - https://archlinuxarm.org/forum/viewtopic.php?t=5374

- https://arstechnica.com/civis/threads/armv7-v7a-v7t-v7k-v7s%E2%80%A6%C2%A0whats-the-difference.1182615/
- https://web.archive.org/web/20131219011024/https://www.keil.com/support/man/docs/armasmref/armasmref_cjafgdih.htm

```text
# Ubuntu RPi
uname -srm
Linux 5.4.0-1098-raspi aarch64
```

```text
# RPi 3 Legacy
uname -srm
# Linux 6.1.21-v7+ armv7l

cat /etc/issue
# Raspbian GNU/Linux 11 \n \l
```

```sh
dpkg --print-architecture
# arm64

sudo dpkg --add-architecture armhf
dpkg --print-foreign-architectures
# armhf

# fetch armhf stuff
sudo apt update
sudo apt-get install libc6:armhf
```

# Notes

Re: https://github.com/webinstall/webi-installers/issues/755, more particularly https://github.com/webinstall/webi-installers/issues/755#issuecomment-1807603923

Matches
[...]

PR #765, opened 2023-11-16
fix+lint release targets

coolaj86:
While working on #756 I discovered
- erroneous releases that cannot be matched generically
- some version numbers that don't match the file path

This updates those release files to produce clean outputs.


jojobyte (approved):
Looks good

coolaj86:
Note: I had to put a hotfix into prod so that the default release is `amd64` for flutter until #756 is ready.

PR #771, opened 2023-11-18
ref: rename to package-install.tpl.sh to match namespace style

coolaj86:
- rename `package-install.tpl.sh` and `package-install.tpl.ps1`
- update direct references
- update paired variables \
  (`-bootstrap` to `-install` to avoid confusion)
- use vim-style `b_` prefix for local vars \
  (this will eventually replace all `my_` vars)


jojobyte (approved):
Do it.

PR #777, opened 2023-11-22
fix: split non-portable function.env => function.sh, function.fish

coolaj86:
`alias` and `export` syntax seem to be common among all shells, but function syntax is different between them.


HacDan (approved):
LGTM!

PR #785, opened 2023-12-12
fix(releases): bump github timeout to 15s, fix file write race condition

coolaj86:
The GitHub source releases can be surprisingly slow, so we bump the timeout from 5s to 15s.


HacDan (approved):
LGTM

PR #786, opened 2023-12-12
fix OS detection when `uname -o` is not supported + typo fix in output

coolaj86:
Alpine (musl) and some BSDs don't always support `uname -o` (Operating System), which is required on Linux to distinguish between GNU and Bionic (Android).

This adds some feature detection for that.


HacDan (approved):
LGTM!

PR #787, opened 2023-12-12
ref: disambiguate names (releases => installers)

coolaj86:
A project
- has releases
  - which has target builds
    - that ship in packages

This is a step towards not mixing up the terminology as much.


HacDan (approved):
LGTM.

PR #652, opened 2023-09-25
add grype + cheatsheet

GreyOwl123:
PR for Grype https://github.com/webinstall/webi-installers/issues/540

PR #809, opened 2024-01-20
feat: added support for anchore/syft, an SBOM utility

joshsleeper:
## changes

- added support for `anchore/syft`, an SBOM utility
    - note that I don't have an easy way to test the `powershell` install script, so hopefully there are automated tests to vet that a bit 🙏 

## ref

- closes <https://github.com/webinstall/webi-installer-requests/issues/25>

### DIFF
--- a/syft/README.md
+++ b/syft/README.md
@@ -1,0 +1,37 @@
+---
+title: Syft
+homepage: https://github.com/anchore/syft/
+tagline: |
+  Syft is a CLI tool and library for generating a Software Bill of Materials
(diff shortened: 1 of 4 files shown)


joshsleeper:
hey @coolaj86, sorry to tag you directly, but figured since this has sat idle for a good while and you reviewed + approved the `grype` MR #652 you might be amenable to doing the same for this~

`syft` is both a standalone CLI tool and literally the internal SBOM tool that `grype` is built on top of

PR #810, opened 2024-02-03
Rename install.ps1 to install.ps1t

Muhammadamjadm:


coolaj86:
What's the reason for using `ps1t`?

Issue #813, opened 2024-02-19
Improve development experience


virzak:
I'm trying to contribute to the repo, but was unable to find a way to do that.

Here are the steps I took:

- Open the repository in code spaces
- Make changes

How do I test before committing?

I tried to start [Live Preview](https://marketplace.visualstudio.com/items?itemName=ms-vscode.live-server). I then changed the URL to `curl -sS https://super-duper-rotary-phone-p66x959p45397g-3000.app.github.dev/shfmt | sh`, but the result I got back was HTML.

There is some sort of url handler, but I have no idea how to invoke it.

The next thing I tried was `yarn start` and `npm start` hoping it would be easy to start a server with proper URL handling, but that was missing also.

Then i read more documentation and after realizing that `packages` were renamed to `webi-installers` ran the following:

```
$ node _webi/test.js ./pwsh
node:internal/modules/cjs/loader:1147
  throw err;
  ^

Error: Cannot find module './build-classifier/host-targets.js'
Require stack:
- /workspaces/webi-installers/_webi/builds-cacher.js
- /workspaces/webi-installers/_webi/builds.js
- /workspaces/webi-installers/_webi/test.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at Object.<anonymous> (/workspaces/webi-installers/_webi/builds-cacher.js:8:19)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Module.require (node:internal/modules/cjs/loader:1235:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/workspaces/webi-installers/_webi/builds-cacher.js',
    '/workspaces/webi-installers/_webi/builds.js',
    '/workspaces/webi-installers/_webi/test.js'
  ]
}
```

The clone command needs to include submodules. Alternatively they should be brought in later if, for example, git clone was donw by codespaces

```sh
git submodule update --init
```

On one of my machines I've been getting:

```
[INFO] getLatestBuilds: kubens
[...]

PR #815, opened 2024-02-19
Use full path to call pwsh, since it is not in the path yet

### LINKED ISSUE #814: Can't install pwsh
## What were you trying to install (or what else went wrong)?

pwsh

## What exactly did you do?

```
curl -sS https://webi.sh/pwsh@7.4.1 | sh
```

## What went wrong?

You should be able to run the same using `docker` if you don't have `podman`

```
🚀  podman run --rm catthehacker/ubuntu:act-latest sh -c "curl -sS https://webi.sh/pwsh@7.4.1 | sh"


>>> Welcome to Webi! - modern tools, instant installs.  <<<
    We expect your experience to be absolutely perfect!

    Success? Star it!   https://github.com/webinstall/webi-installers
    Problem? Report it: https://github.com/webinstall/webi-installers/issues
                        (your system is GNU/Linux/x86_64 with libc & curl+wget)

Bootstrapping Webi
    Downloading https://webi.sh/packages/webi/webi.sh
        to ~/.local/bin/webi
    Running ~/.local/bin/webi pwsh@7.4.1

Installing pwsh ...
    Found  ~/.local/bin
    Downloading pwsh from
      https://github.com/PowerShell/PowerShell/releases/download/v7.4.1/powershell-7.4.1-linux-x64.tar.gz
    Saved as ~/Downloads/webi/pwsh/7.4.1/powershell-7.4.1-linux-x64.tar.gz
    Extracting ~/Downloads/webi/pwsh/7.4.1/powershell-7.4.1-linux-x64.tar.gz
    Installing to ~/.local/opt/pwsh-v7.4.1/pwsh
    Initializing ~/.config/envman/
    Edit ~/.profile to source ~/.config/envman/load.sh
    Edit ~/.bashrc to source ~/.config/envman/load.sh
Installed 'pwsh' at /root/.local/opt/pwsh
pwsh@7.4.1-install.sh: 470: pwsh: not found
```

## Which OS did you try on?

- [ ] Mac
- [ ] Windows
- [x] Linux

## What type of computer (i.e. laptop, desktop, Raspberry Pi)?

- [x] Laptop or Desktop (amd64 / x86-64)
- [ ] Raspberry Pi (ARM-64)
- [ ] Other (please specify, if you know)

## Possible explanation

Since pwsh is not in $PATH yet, it needs to be called fully qualified.
https://github.com/webinstall/webi-installers/blob/c431ce0b22dd4a577af1be4f0ebf5c0a22c689ef/pwsh/install.sh#L49

virzak:
Resolves #814 

Tried to test, but failed due to #813

### DIFF
--- a/pwsh/install.sh
+++ b/pwsh/install.sh
@@ -45,7 +45,9 @@ __init_pwsh() {
     }
 
     pkg_done_message() {
-        echo "Installed 'pwsh' at $pkg_dst"
+        # We print the version here to ensure the install completed without
+        # errors - no missing libraries, not incompatible arch, etc
+        echo ""
         "$pkg_dst_cmd" -V
     }
 }
--- a/pwsh/install.sh
+++ b/pwsh/install.sh
@@ -46,7 +46,7 @@ __init_pwsh() {
 
     pkg_done_message() {
         echo "Installed 'pwsh' at $pkg_dst"
-        pwsh -V
+        "$pkg_dst_cmd" -V
     }
 }
 


coolaj86:
I'm trying to remember why I had the custom `pkg_done_message` here at all.

My guess is that it's to surface errors when dependencies (libraries) are missing, or the wrong binary is selected for the target machine.

Making a few modifications:

- `"$pkg_dst_cmd"` must be quoted, otherwise Macs on Windows networks get errors when the Active Directory domain creates a space in the user's home directory. For example: `/Users/John Smith/.local/opt/pwsh/pwsh`
- Nixing the "installed to" since we already have that in the new install script

coolaj86 (approved):
Thank you!

coolaj86:
I was pretty sure that the command should be in the PATH by virtue of the installer's processing having it in its PATH.

So... I don't think this actually fixed anything. I think the real problem was Microsoft's change to the package internals.

PR #816, opened 2024-02-19
Try testing in container

virzak:


virzak:
Hi @coolaj86,

Could you please approve the workflow?

virzak:
Hi @coolaj86. 
Hi @coolaj86,

Could you please approve the workflow?

coolaj86:
Hi @virzak,

Thanks for looking into this. I'm now available to focus on Webi during the next 2-3 weeks. If you're also available, I'll have a quick turn around time to work with you on this.

virzak:
@coolaj86 let's do it!

coolaj86:
It looks like what you have here is simply creating a container, copying a curl binary into it, and then running webi on the fresh system to see if it installs `pwsh`.

Where do you want to take it?

I think my ideal would be something like:
- not run on every push, but must be run before any `ref:` or `feat:` PR can be accepted \
  (GitHub Actions is distractingly slow to have run on each commit)
- track which things are expected to not work on Alpine, or macOS, or Windows
- check the `-V` / `--version` of each thing as the last step

I also have a High-Availability Proxmox cluster where I can create LXCs within a matter of a few seconds (I could even preallocate some), and delete them just as easily. That could be much faster than GitHub Actions, and give access to any environment... but also comes at the cost of development.

virzak:
This was just a first investigation step. The intention was to expose #814 during build.

Since this workflow was just recently approved I didn't know how long this would take. It also looks like curl needs more permissions, so I'll try to modify it.

Definitely if it takes too long, perhaps there needs to be a different strategy.

virzak:
After fixing the permission issue, the container seems to be taking about 5-6 seconds.
It currently fails because it uses the production webi, which still has the #814 bug, so it needs to be adjusted to use webi produced in this build.

coolaj86:
Try installing 7.3 instead of latest / 7.4

PR #819, opened 2024-03-01
docs(README.md): Update readme to reflect repo name change

HacDan:
Updated README.md from `packages` to `webi-installers` to be in line with current project structure.

If you want to skip the PR process for this, I completely understand. This came up in issue #813.


y0rune:
I agree with it! Due to exist file `.gitmodule` it should be add into a `README.md`

PR #841, opened 2024-05-16
fix: update ssh-pubkey for modern 3072+ bit and ed25519 keys

coolaj86:
- generates RSA 4096 bit keys when no key is available (old behavior was 2048 bit)
- supports ED25519 and ECDSA keys
- will print a warning for RSA keys less than 3072 bits

The relevant commit is https://github.com/webinstall/webi-installers/pull/841/commits/6172fba55fd2303912cd07289b342d7e5d6f46ec

The other 4 commits are just some cleanup.

Re:
- https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent


ssh-pubkey/ssh-pubkey
```diff
@@ -2,43 +2,99 @@
...
+        ssh-keygen -y -f ~/.ssh/id_rsa > ~/.ssh/id_rsa.pub
     fi
 
     # TODO use the comment (if any) for the name of the file
     echo >&2 ""
     #shellcheck disable=SC2088
-    echo >&2 "~/Downloads/id_rsa.$(whoami).pub":
+    echo >&2 "~/Downloads/id_${my_keytype}.$(whoami).pub":
     echo >&2 ""
-    rm -f "$HOME/Downloads/id_rsa.$(whoami).pub"
```
jojobyte, line 38: like this is still `$HOME`

ssh-pubkey/ssh-pubkey
```diff
@@ -2,43 +2,99 @@
...
+    # shellcheck disable=SC2016
+    echo '    ssh-keygen -t rsa -b 4096 -f ~/.ssh/id_rsa -q -N ""'
+    echo "    echo 'Host *
+        IdentityFile ~/.ssh/id_rsa
+        IdentityFile ~/.ssh/id_rsa.${my_ts}.bak' >> ~/.ssh/config"
+); }
+
 main() {
 
-    if [ ! -d "$HOME/.ssh" ]; then
```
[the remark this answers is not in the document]
coolaj86: Using `~/` where the paths don't need to be quoted, but leaving "$HOME" where it _does_ need to be quoted - because quoting "~/" escapes it.

ssh-pubkey/ssh-pubkey
coolaj86: It has to be quoted for the "$(whoami)". Options are:

1. `rm -f "$HOME/Downloads/id_rsa.$(whoami).pub"`
2. `rm -f ~/Downloads/id_rsa."$(whoami)".pub`
3. `rm -f ~/"Downloads/id_rsa.$(whoami).pub"`

Thoughts?

jojobyte (approved):

Issue #854, opened 2024-06-05
Simple refactor: find-and-replace `whoami` with `id -u -n` in scripts


coolaj86:
`whoami` is generally available on all Unix-y systems, but `id -u -n` is the POSIX standard with guaranteed behavior.

This is literally as simple as doing a find-and-replace across the repo (such as with [`sd`](https://webinstall.dev/sd), or any code editor tool).

Finding the files with `whoami`:

```sh
git clone https://github.com/webinstall/webi-installers ./webi-installers
pushd webi-installers/
rg 'whoami' || grep -r 'whoami' .
```

For reference: https://pubs.opengroup.org/onlinepubs/009695399/utilities/id.html

shuchitpant:
Hi AJ,

I've raised a PR (#915 ), that addresses this issue, kindly go through this and let me know if there are any changes.

coolaj86:
Thank you!

PR #857, opened 2024-06-10
fix(#854)Update whoami to id -u -n in scripts

HacDan:
I couldn't find any others, but my `rg` foo isn't that strong.


coolaj86:
Looks about right. Will review and pull in later today or tomorrow.

PR #890, opened 2024-09-11
fix(linux-gnu): vupdate to 1.0.0: respect explicit 'gnu'

coolaj86:
Brings in this change:
https://github.com/webinstall/webi-build-classifier/pull/11/files

```js
    // See <https://github.com/webinstall/webi-build-classifier/pull/11>
    if (target.os === 'linux') {
      if (target.libc === 'none') {
        let isGnuOnly = terms.includes('gnu');
        if (isGnuOnly) {
          target.libc = 'gnu';
        }
      }
    }
```

### DIFF
--- a/_webi/builds-cacher.js
+++ b/_webi/builds-cacher.js
@@ -189,9 +189,10 @@ BuildsCacher.create = function ({ ALL_TERMS, installers, caches }) {
   }
 
   let bc = {};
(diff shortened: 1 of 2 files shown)


ryanburnette (approved):

coolaj86:
Brought in with https://github.com/webinstall/webi-installers/pull/891

PR #895, opened 2024-09-14
fix(runzip): use .getDistributables()

coolaj86:
runzip was added before the ` .getDistributables()`.

This updates `runzip` to use the correct function.

This also cleans up `.gitignore` to reduce clutter on `beta`.


jojobyte (approved):
Cowabunga!

Issue #879, opened 2024-08-21
[Bug] `bun` should prefer `bun-baseline` 


coolaj86:
- [x] `linux` (vm, container, NOT desktop)
 - NOT `macos`
 - NOT `windows`

Containers often play to the lowest common denominator for CPU virtualization. That leads to this:

```text
Bun v1.1.25 (fe62a614) Linux x64
fish: Job 1, 'bun' terminated by signal SIGILL (Illegal instruction)
```

We either need to either figure out what specific CPU features pertain to "profile" vs "" vs "baseline" and detect them, or just serve baseline, for Linux at least.

**Update** here's how we find out:

```sh
cat /proc/cpuinfo | grep avx2
```

More: https://github.com/oven-sh/bun/issues/7607#issuecomment-1854855079

Jarred-Sumner:
It should auto-detect the CPU platform instead of always choosing baseline. We have a bunch of SIMD optimizations that are disabled on the baseline build

coolaj86:
@Jarred-Sumner We don't yet have a standard method for detecting cpu-level features, but I have some ideas for how to implement it when I have time.

coolaj86:
Probably not 100% accurate, but here's a starter: \
(however, I'm not sure how far down this rabbit hole I really want to go - it could turn out to be a lot of complex bloat)

> To detect specific CPU features that can help identify AMD processor generations (such as AMDv2, AMDv3, and AMDv4), you can use various commands and utilities depending on the operating system. Below is a matrix for Linux, Windows, macOS, FreeBSD, OpenBSD, and NetBSD.

| **Operating System** | **Command for AMDv2 (AVX)**               | **Command for AMDv3 (AVX2)**               | **Command for AMDv4 (AVX-512)**            |
|----------------------|-------------------------------------------|--------------------------------------------|--------------------------------------------|
| **Linux**            | `cat /proc/cpuinfo \| grep avx`           | `cat /proc/cpuinfo \| grep avx2`           | `cat /proc/cpuinfo \| grep avx512`         |
| **Windows**          | `wmic cpu get name, Caption \| findstr /C:"AVX"` | `wmic cpu get name, Caption \| findstr /C:"AVX2"` | `wmic cpu get name, Caption \| findstr /C:"AVX-512"` |
[...]

PR #897, opened 2024-09-14
fix(ollama): classify 'rocm' as its own cpu

coolaj86:
As per @jojobyte report of the Ollama CPU mismatch.

See also:
- https://github.com/webinstall/webi-build-classifier/pull/15
- https://github.com/webinstall/webi-installers/issues/879
- https://github.com/golang/go/issues/45453


jojobyte (approved):
Looks good.

coolaj86:
Looks can be deceiving.

But after much trial and error (meaning manual testing), it's now confirmed working on Linux, macOS, and Windows.

PR #923, opened 2024-12-08
Comply with XDG Base Dir Specs

pothi:
To comply with [XDG Base Dir Specs](https://specifications.freedesktop.org/basedir-spec/latest/index.html). As of Vim [Patch 9.1.327](https://github.com/vim/vim/commit/c9df1fb35a1866901c32df37dd39c8b39dbdb64a), Vim started supporting XDG Base Dir specs. We don't need to use Vim version 9.1.327 (or higher) to use `viminfo` and changes made by this pull request.

### DIFF
--- a/vim-viminfo/README.md
+++ b/vim-viminfo/README.md
@@ -13,13 +13,14 @@ These are the files / directories that are created and/or modified with this
 install:
 
 ```text
-~/.vim/plugins/viminfo.vim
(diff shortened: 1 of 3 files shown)


coolaj86:
It sounds like this is something that should be circled back around to when vim 9.1 is widely deployed across the ecosystem?

PR #934, opened 2024-12-15
fix(api): 'alias' projects (symlinks) should be resolved before checking for 'selfhosted'

### LINKED ISSUE #933: problem with vim-gui installer
>>> Welcome to Webi! - modern tools, instant installs.  <<<
    We expect your experience to be absolutely perfect!

    Success? Star it!   https://github.com/webinstall/webi-installers
    Problem? Report it: https://github.com/webinstall/webi-installers/issues
                        (your system is Darwin/arm64 with libc & curl)

Bootstrapping Webi
    Found ~/.local/bin/webi
    Running ~/.local/bin/webi vim-mouse@stable

curl: (56) The requested URL returned error: 500
error fetching 'https://webi.sh/api/installers/vim-mouse@stable.sh?formats=tar,exe,zip,xz,git,dmg,pkg'

coolaj86:
fixes #933 

Projects are marked as an `alias` if they're a symlink, or if `alias: ` is matched in the first 1k of of the `README.md`.

Excluding possible errors, the resolved projects are either `valid` (having a `releases.js`) or `selfhosted` (having an `install.sh` and local files committed to Webi, no remote releases).

Up to this point `alias` projects have simply been cached twice. However, `vim-mouse` is a symlink alias of `vim-gui`, which has no `releases.js` and was therefore failing module resolution (as it didn't match `selfhosted`).

This change resolves `alias` projects before checking if they're `valid` or `selfhosted`.

### DIFF
--- a/_webi/serve-installer.js
+++ b/_webi/serve-installer.js
@@ -70,17 +70,19 @@ InstallerServer.helper = async function ({
 
   console.log(`dbg: Get Project Installer Type for '${projectName}':`);
   let proj = await Builds.getProjectType(projectName);
-  console.log(proj);
+  if (proj.type === 'alias') {
+    console.log(`dbg: alias`, proj);
+    projectName = proj.detail;
+    proj = await Builds.getProjectType(projectName); // an alias should never resolve to an alias
+  }
+  console.log(`dbg: proj`, proj);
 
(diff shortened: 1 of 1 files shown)


jojobyte (approved):
Looks okay but do you want the console debugs still in there?

Issue #941, opened 2024-12-24
[Bug] `gh`: Incorrect windows ARCH parsing


pyhedgehog:
## Description

On windows [gh.tab](https://webinstall.dev/api/releases/gh.tab?channel=stable&limit=6&pretty=true&os=windows) you can see that both `gh_*_windows_386.msi` and `gh_*_windows_amd64.msi` has `ARCH=amd64`.

## Expected

It's expected that `*_386.*` marked as `x86`.

## Proposal

https://github.com/webinstall/webi-installers/blob/976602236b434be17c218b39387496d99216a40b/_webi/normalize.js#L60
Here regexp should be changed to:
`/(\b|_|amd|(dar)?win(dows)?|mac(os)?|linux|osx|x)(86|32|386)([_\-]?bit)?(\b|_)/i`.

coolaj86:
Thanks for the report and the fix. I'll update this tomorrow morning.

PR #942, opened 2024-12-24
Fixes GH-941

pyhedgehog:


pyhedgehog:
@coolaj86
When "tomorrow" will happens? ;-)

coolaj86 (approved):
Changing this regexp is so scary (it'll break something, I'm sure)... but we'll do it!

The tab API is no longer used internally - it's been replaced by the build-classifier. But until I have time to update the tab API I suppose we might as well fix this.

coolaj86:
> @coolaj86 When "tomorrow" will happens? ;-)

Today!

PR #968, opened 2025-02-19
fix(koji): update binary path, docs

cococonscious:
The release process of koji has been updated a while ago, resulting in the binary not being nested in the archive anymore.

Fixes https://github.com/cococonscious/koji/issues/137

### DIFF
--- a/koji/install.sh
+++ b/koji/install.sh
@@ -19,13 +19,13 @@ __init_koji() {
         # ~/.local/opt/koji-v1.5.0/bin
         mkdir -p "$(dirname "$pkg_src_cmd")"
 
-        # mv ./koji-*/koji ~/.local/opt/koji-v1.5.0/bin/koji
-        mv ./koji-*/koji "$pkg_src_cmd"
+        # mv koji ~/.local/opt/koji-v1.5.0/bin/koji
+        mv koji "$pkg_src_cmd"
     }
 
     # pkg_get_current_version is recommended, but (soon) not required
     pkg_get_current_version() {
-        # 'koji version' has output in this format:
+        # 'koji --version' has output in this format:
         #       koji 1.5.0
         # This trims it down to just the version number:
         #       1.5.0
--- a/koji/install.sh
+++ b/koji/install.sh
@@ -19,8 +19,19 @@ __init_koji() {
         # ~/.local/opt/koji-v1.5.0/bin
         mkdir -p "$(dirname "$pkg_src_cmd")"
 
-        # mv koji ~/.local/opt/koji-v1.5.0/bin/koji
-        mv koji "$pkg_src_cmd"
+        if test -f ./koji; then
+            # mv koji ~/.local/opt/koji-v1.5.0/bin/koji
+            mv ./koji "$pkg_src_cmd"
+        elif test -e ./koji-*/koji; then
+            # mv koji-1.4.0/koji ~/.local/opt/koji-v1.4.0/bin/koji
+            mv ./koji-*/koji "$pkg_src_cmd"
+        elif test -e ./koji-*; then
+            # mv koji-linux-amd64 ~/.local/opt/koji-v1.2.0/bin/koji
+            mv ./koji-* "$pkg_src_cmd"
+        else
+            echo >&2 "failed to find 'koji' exectutable"
+            return 1
+        fi
     }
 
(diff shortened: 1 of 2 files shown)


cococonscious:
By the way, sorry to be so upfront about it, but the developer experience is terrible. 

The `_webi/test.js` script isn't functional, nor helpful with debug/error messages and the usage guide is just straight up wrong (e.g. adding the `--debug` flag triggers the usage help error). From another issue I got the idea that I could just self host it locally real quick, but there's no guide to do so, nor any container I could quickly spin up.

Perhaps if I spent more time debugging I could fix the issue myself somehow, but that just shouldn't be expected of people wanting to test simple edits on their scripts (especially since simplicity seems to be a core goal of this project).

By no means am I trying to diminish your work, webi is great, but this part is just not that great.

coolaj86:
> By the way, sorry to be so upfront about it, but the developer experience is terrible.

I agree and I take no offense.

image
(me)

["Simplicity is complicated"](https://www.youtube.com/watch?v=rFejpH_tAHM) and I've taken some wrong turns on the way to get the project to where it is today.

If you're free to schedule some time and work through the testing process together, that would be helpful to me to have the motivation to focus on it for a few hours. If so, what timezone are you in?

koji/install.sh
```diff
@@ -19,13 +19,13 @@ __init_koji() {
...
-        # mv ./koji-*/koji ~/.local/opt/koji-v1.5.0/bin/koji
-        mv ./koji-*/koji "$pkg_src_cmd"
+        # mv koji ~/.local/opt/koji-v1.5.0/bin/koji
+        mv koji "$pkg_src_cmd"
     }
 
     # pkg_get_current_version is recommended, but (soon) not required
     pkg_get_current_version() {
-        # 'koji version' has output in this format:
+        # 'koji --version' has output in this format:
```
coolaj86, line 28: `--version`, `-V`, and `version` should all be valid

coolaj86 (approved):
If `koji version` worked before, I'd suggest not breaking that, as scripts may rely on it.

In general, `-V` is the expected POSIX flag, `--version` is the expected GNU flag, and `version` is the expected Go subcommand, so supporting all 3 is a good idea.

I'm going to add a little shim to this so that it doesn't break old versions.

cococonscious:
> If `koji version` worked before, I'd suggest not breaking that, as scripts may rely on it.
> 
> In general, `-V` is the expected POSIX flag, `--version` is the expected GNU flag, and `version` is the expected Go subcommand, so supporting all 3 is a good idea.
> 
> I'm going to add a little shim to this so that it doesn't break old versions.

1. The `version` subcommand hasn't ever existed on koji. I can add it if you want me to but I'd prefer to fix the install script first.
2. Your fix for backwards compatibility is nice, thank you, but it's actually prior to version 3.0.0, not 1.4

cococonscious:
> If you're free to schedule some time and work through the testing process together, that would be helpful to me to have the motivation to focus on it for a few hours. If so, what timezone are you in?

I'd like to, but I'm very busy in the upcoming weeks so I can't really spend any time on this. If you want to wait 1-2 months feel free to create an issue and link me in it.

coolaj86:
I tested it with a few different versions pushed to main.

Go ahead and test it on your end and see if it's working as you expect.

coolaj86:
One thing that I'm a little confused about is that it's defaulting to koji 1.4 for Linux, but 3.2 for Mac.

Is the linux version marked "beta" instead of stable? Or something else that might cause the version auto detection to be overwritten?

Is there anything that comes to mind for you as to why that might be?

cococonscious:
Hm, these are the differences from v3.2 to v1.4:
- Tag now contains the "v" prefix
- "osx" is now "darwin"
- Archive name doesn't contain version anymore
- There's no more musl build for linux, only gnu

Any ideas? Perhaps you system is using musl?

coolaj86:
> Perhaps your system is using musl

Webi prefers musl builds because they typically don't have any dependencies, whereas gnu builds are guaranteed to have dependencies which have conflicts between versions of Linux.

Why drop musl builds when they're so much more reliable?

cococonscious:
> Why drop musl builds when they're so much more reliable?

I quite simply forgot. I'll add them to future releases of koji.

PR #971, opened 2025-03-06
Update README.md

bry-val:
deprecated path to deno example


coolaj86:
The .com always wins...

Thanks for the update!

coolaj86 (approved):

coolaj86 (approved):

PR #1025, opened 2025-10-11
fix(rg): Handle missing rg command in install.sh

loveyou001:
The `pkg_get_current_version` function in `rg/install.sh` would fail if the `rg` command was not found in the `PATH`. This change adds a check to ensure the command exists before attempting to execute it, preventing the installation script from failing unexpectedly.

### DIFF
--- a/rg/install.sh
+++ b/rg/install.sh
@@ -33,6 +33,10 @@ __init_rg() {
 
     # pkg_get_current_version is recommended, but (soon) not required
     pkg_get_current_version() {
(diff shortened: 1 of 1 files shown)


loveyou001:
.....

detox-24 (approved):

coolaj86:
@detox-24 This was Ai-generated spam. If you look at the PR it's garbage - solving a non existent problem (if `rg` was just installed, then it will exist).

detox-24:
Oops! my bad!

PR #1048, opened 2026-03-08
fix(bun): prefer baseline linux releases (fix #879)

### LINKED ISSUE #879: [Bug] `bun` should prefer `bun-baseline`

Tori0419:
Fixes #879

## Summary

- keep Bun's `-baseline` Linux assets instead of filtering them out
- prefer the baseline build over the default build for the same Linux x64 target
- keep debug/profile builds excluded
- add a narrow regression test for release ordering

## Why

`bun` can select a non-baseline Linux binary that crashes with `SIGILL` on older/container CPUs.
This change makes the release list prefer the baseline variant when both assets exist for the same target.

## Testing

- added `bun/releases.test.js` covering:
  - baseline preferred over default Linux x64 assets
  - non-Linux asset order unchanged

## Notes

I could not run the repo's Node-based lint/test commands locally in this environment because `node`/`npm` were unavailable.


coolaj86:
Weird, I thought I had fixed this years ago. Looks good. Testing now.

coolaj86 (approved):
The sort strategy alone didn't work, but I changed it to just use baseline when baseline is available and that solves the problem.

coolaj86:
@detox-24 This was also kinda Ai slop (it didn't actually understand the problem to fix it correctly). However, it was at least attempting to address an actual problem.

detox-24:
Ah! I gotta keep a sharper eye!

PR #1016, opened 2025-08-11
Add tlrc package - fast tldr client written in Rust

copyleftdev:
- Add tlrc/releases.js using GitHub API integration
- Add tlrc/install.sh with proper error handling and version detection
- Add tlrc/README.md with comprehensive documentation and examples
- Cross-platform support: Linux, macOS, Windows
- Follows all webi-installers best practices and formatting standards

tlrc is the official Rust-based tldr client that provides community-maintained help pages for command-line tools with practical examples instead of lengthy man pages. This package adds support for installing tlrc v1.0.0 through v1.11.1 with automatic version detection and proper PATH management.

PR #1060, opened 2026-03-14
feat: Add tlrc package - fast tldr client written in Rust

detox-24:
Originally authored by @copyleftdev as PR #1016. 

Tagged ```gnu``` and ```musl``` builds of releases in releases.js
The tlrc package still uses ```tldr``` as the command name. Updated the same in install.sh

~~TODO: Update README.md~~

Updated README.md with correct command name. (Ref: https://tldr.sh/tlrc/)

### DIFF
--- a/tlrc/README.md
+++ b/tlrc/README.md
@@ -1,0 +1,71 @@
+---
+title: tlrc
+homepage: https://github.com/tldr-pages/tlrc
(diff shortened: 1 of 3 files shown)


coolaj86:
As soon as I can deploy the new version I'll be bringing this in.

Issue #1056, opened 2026-03-12
We need an Agent Skill for creating an Agent Skill (yep)


coolaj86:
We're going to be creating a lot of Agent Skills because, in my testing so far, they're incredibly useful. They can help an agent use a tool, and the can also help an agent help a person use a tool. And they're more general, more useful, and faster to develop than MCP.

But, the irony of https://agentskills.io/specification is that they don't yet have an agent skill for creating agent skills... so we'll need to create one.

This is probably pretty easy in the sense that we can probably just give the URL and the repo to an agent and ask it to create a skill with the documentation and then do some verification that it can use it.

Re #1041, #1055, https://github.com/agentskills/agentskills/issues/232

PR #1061, opened 2026-03-17
Add Agent Skill for generating new Agent Skills

kodurusravani34:
Added an Agent Skill that generates new Agent Skills from documentation or repository inputs. Helps simplify and scale skill creation.

### DIFF
--- a/agent-customization/agent-skill-generator.md
+++ b/agent-customization/agent-skill-generator.md
@@ -1,0 +1,94 @@
+# Agent Skill: Create Agent Skill
+
+## Description
+Generate a new Agent Skill from a documentation URL or GitHub repository.  
+This skill automates the creation of reusable Agent Skills by analyzing source material and structuring it according to the Agent Skills specification.
+
+---
+
+## Input
+- **documentation_url** (required): URL to documentation (e.g., https://agentskills.io/specification)  
+- **repository_url** (optional): GitHub repository link  
+- **context** (optional): Additional notes about the tool or use-case  
+
+---
+
+## Output
+- Skill Name  
+- Description  
+- Inputs  
+- Outputs  
+- Step-by-step Instructions  
+- Example Usage  
+
+---
+
+## Instructions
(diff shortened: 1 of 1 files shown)


David-Bosnic:
~~This has no use case for webi and would just adds a markdown file that pretty much does nothing.~~
Refer to @detox-24 comment and mine after.
Closing

detox-24:
Hello @kodurusravani34! Your idea is in the right place. Might I suggest some changes/improvements to this PR.

1. Wrong file structure

Agent Skills follow the [agentskills.io spec](https://agentskills.io/specification), which requires a named directory containing a SKILL.md file, not a standalone .md file. The correct structure looks something close to this:
```bash
_skills/agent-skill-generator/
├── SKILL.md
└── references/        (optional but encouraged)
    └── REFERENCE.md
```
Your file was placed at agent-customization/agent-skill-generator.md which does not follow the spec. Skills in this repo live under _skills/. For a real example of what this looks like in practice, check out PR #1059 by @coolaj86, that is a good reference for structure and depth.

2. Missing YAML frontmatter

The SKILL.md must start with a YAML frontmatter block. At minimum:

```yaml
---
name: agent-skill-generator
description: Generates a new Agent Skill for any tool by analyzing its documentation URL or repository. Use when asked to create or scaffold a new Agent Skill.
---
```

3. Add a verification step

Issue #1056 specifically calls this out. Once the skill generates a new Agent Skill, there should be a way to confirm the output actually works, not just that it looks right. Even a short checklist would do: does the generated skill have valid frontmatter? Can an agent follow the instructions end to end? Does it cover both agent use and human-via-agent use?


@David-Bosnic, I think this one deserves a chance to update, thoughts?

David-Bosnic:
Fair. I did shoot the gun.
@kodurusravani34
Open a new pr with a better description and follow @detox-24 suggestions. 
A better understanding of what and why will help in the future when people look back into the PR
even though the preposed merge is small.

PR #1075, opened 2026-05-07
ref(builds-cacher): cache-only Node server + API libc-filter fix

coolaj86:
## Summary

- Node server reads exclusively from `~/.cache/webi/legacy/<name>.json`
- Removes every code path that fetched upstream or wrote to the cache
- Webicached (Go daemon) is now the sole writer; this server is a thin reader
- Fix `/api/releases` filtering musl entries when the caller didn't pin a libc

## What changed

`_webi/builds-cacher.js`:
- Resolves cache files via `Os.homedir() + '/.cache/webi/legacy/'` instead of the `cacheDir`/`caches` argument.
- Drops `getLatestBuilds` / `getLatestBuildsInner` — they `require()`d per-package `releases.js`, fetched upstream, and wrote `<yyyy-mm>/<name>.json` + `.updated.txt` to disk.
- Drops the `process.nextTick` stale-refresh hook in `_doGetPackages`. Cold reads return what's on disk; if the file is missing, return empty meta instead of fetching.
- Drops `freshenRandomPackage` and supporting state (`bc._staleNames`, `bc._freshenTimeout`, `bc._staleAge`). The hourly background freshener was competing with webicached for the same files.
- In `getProjectTypeByEntry`, decides selfhosted vs valid by probing for the cache file rather than `require()`-ing `releases.js`.

`_webi/transform-releases.js`:
- Drops `Releases.get` and the `_normalize` import.
- Replaces `getCachedReleases`'s fetch+race+stale-age machinery with a single `Fs.readFile` of `~/.cache/webi/legacy/<pkg>.json`.
- **`fix(api)`:** the libc filter ran unconditionally — even when the caller didn't pin a libc, the default `libc='libc'` from `serve-releases.js` was triggering it and stripping every `libc=musl` entry from listings. Treat the catch-all `'libc'` value as 'no preference' so the website's release table and the `WEBI_RELEASES` probe see what the cache actually contains. Specific `?libc=musl` / `?libc=gnu` queries still filter exactly as before.

`_webi/build-classifier`:
- Submodule pin bumped to v1.0.3 (the `Lexver.parsePrefix` zero-padding fix for `<pkg>@<exact-version-with-suffix>` queries).

No public signatures change. Per-package `releases.js`, `_common/*.js`, and `_webi/normalize.js` still exist on disk but are no longer reachable from the request path (cleanup follows in a separate PR).

## Test plan (verified against `https://webinstall.dev/`)
- [x] `~/.cache/webi/legacy/` exists on prod and is populated (100 packages).
- [x] `curl https://webinstall.dev/api/releases/go.json?limit=1` returns valid version (1.26.3, not `0.0.0`).
- [x] `curl -A 'curl/8.5 Linux x86_64' https://webinstall.dev/node` returns `#!/bin/sh` installer.
- [x] Unfiltered listing returns the non-error shape (libc-fix verified on beta where the cache contains musl-tagged entries: `40 none / 44 gnu / 16 musl`). On prod the cache currently has 0 musl-tagged entries (`50 none / 50 gnu`) because webicached classifies static-musl Rust builds as `libc=none`. The fix is forward-looking once musl-tagged entries land in prod's cache.
- [ ] `?libc=musl` strict-filter test (`curl '…&libc=musl&limit=5'`) is **not reachable** as written: `webinstall.dev/lib/serve-releases.js:37` runs the libc query through `UaDetect.arch()` which mangles non-arch strings to `'error'` before reaching this filter. Separate fix needed in that file (out of scope here).
- [ ] Server-log audit and "no writes under `~/.cache/webi/` from the Node process" are deploy-host checks the maintainer will confirm.

### DIFF
--- a/_webi/builds-cacher.js
+++ b/_webi/builds-cacher.js
@@ -3,13 +3,16 @@
 var BuildsCacher = module.exports;
 
 let Fs = require('node:fs/promises');
+let Os = require('node:os');
 let Path = require('node:path');
+
+let LEGACY_CACHE_DIR = Path.join(Os.homedir(), '.cache/webi/legacy');
 
 let HostTargets = require('./build-classifier/host-targets.js');
 let Lexver = require('./build-classifier/lexver.js');
 let Triplet = require('./build-classifier/triplet.js');
 
-var ALIAS_RE = /^alias: (\w+)$/m;
+var ALIAS_RE = /^alias: ([\w.-]+)$/m;
 
 var LEGACY_ARCH_MAP = {
   '*': 'ANYARCH',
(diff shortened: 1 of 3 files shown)


_webi/builds-cacher.js
```diff
@@ -3,13 +3,16 @@
...
 let Path = require('node:path');
 
+let LEGACY_CACHE_DIR = Path.join(Os.homedir(), '.cache/webi/legacy');
+
 let HostTargets = require('./build-classifier/host-targets.js');
 let Lexver = require('./build-classifier/lexver.js');
 let Triplet = require('./build-classifier/triplet.js');
 
-var ALIAS_RE = /^alias: (\w+)$/m;
+var ALIAS_RE = /^alias: ([\w.-]+)$/m;
```
coolaj86, line 15: re: vim-zig, zig.vim

PR #1099, opened 2026-05-17
fix(deploy): exclude .claude worktrees from releases.conf rsync

coolaj86:
## Summary

- Adds `--exclude='.claude'` to the rsync in `deploy-webicached.sh`
- Without this, worktrees under `.claude/worktrees/` were picked up by the `--include='releases.conf'` rule, syncing ~1300 extra files to the server on every deploy from the main repo root

## Test plan

- [x] Verified rsync now transfers only the expected installer `releases.conf` files


coolaj86:
dumb Ai junk

Issue #1115, opened 2026-07-25
[Suggestion] Making it mandaotry to tag 'AI-assisted' in new PRs


detox-24:
We are starting to receive a good number of PRs for new installers since migrating the architecture. 

And a good amount of it is completely out of tracks and difficult to review. 

I suggest it would be easier to review PRs if the disclosure of AI-Assistance is made compulsory. 

And no, it is not in my interest to completely disregard the PR just because it is AI assisted. It would just help in knowing what we get into beforehand.

coolaj86:
Thanks again for this @detox-24!

PR #1124, opened 2026-08-27
fix(webi): fetch git-source packages without annotated-tag warnings

### LINKED ISSUE #1078: webi_git_clone: cosmetic 'is not a commit' warning for annotated tags
## Summary

`webi_git_clone` in `_webi/package-install.tpl.sh` uses
`git clone --depth=1 --single-branch --branch <tag>` to fetch
git-source packages. When the upstream tag is **annotated** (the
common case for signed releases — `git tag -a` or `git tag` with
`tag.gpgsign = true`), git's shallow protocol returns the tag-object's
SHA rather than a commit SHA, and the clone code path emits:

```
warning: refs/tags/<TAG> <SHA> is not a commit!
```

The clone proceeds correctly (peels through `tag → commit`, working
tree ends up at the right state), so this is purely a diagnostic. But
it shows up in every install for git-source packages whose upstream
publishes annotated tags, and that's the majority of well-maintained
projects.

## Reproduction

Any git-source install whose upstream tag is annotated produces the
warning. The clone succeeds and the install completes.

## Suggested fix

Replace `git clone --depth=1 --branch <ref>` with
`git init + git fetch --depth=1 origin <ref> + git checkout FETCH_HEAD`.
The `fetch` form handles annotated tags, lightweight tags, and
branches uniformly with no warning, and produces the same working
tree.

```sh
mkdir -p "${my_dl}.part"
(
    cd "${my_dl}.part"
    git init --quiet
    git remote add origin "${my_url}"
    git fetch --depth=1 --quiet origin "${WEBI_GIT_TAG}"
    git checkout --quiet FETCH_HEAD
)
```

## Notes

- Functional behavior is unchanged — `clone --branch` and
  `fetch + checkout FETCH_HEAD` both end up at the same commit.
- Visible diff for users is just the absence of the inert warning.
- The warning's appearance can vary by git client version (some
  platforms emit it, some don't). Switching to `fetch` is uniform
  across clients.

## Affected

- `_webi/package-install.tpl.sh` (`webi_git_clone`)

Agi-Asi:
### Problem

`webi_git_clone` uses `git clone --depth=1 --single-branch --branch <tag>`. When the upstream tag is **annotated** (the common case for signed releases — `git tag -a`, or `tag.gpgsign=true`), git's shallow protocol returns the tag object's SHA rather than a commit SHA, and some git versions emit a cosmetic but alarming warning on every install of a git-source package:

```
warning: refs/tags/<TAG> <SHA> is not a commit!
```

The clone itself succeeds (git peels tag → commit correctly), so this is purely diagnostic noise — but it lands in the middle of webi's otherwise clean install output, exactly the kind of "necessary but unimportant step leaking into the experience" the AGENTS.md philosophy says shouldn't happen.

### Fix

Replace the clone with `git init` + `git fetch --depth=1 origin <ref>` + `git checkout FETCH_HEAD`. The fetch form handles **annotated tags, lightweight tags, and branches uniformly**, with no warning and the same resulting working tree.

Details:
- `git init -b main` avoids the default-branch hint on older git configs; the branch name is irrelevant since checkout detaches to FETCH_HEAD anyway.
- `advice.detachedHead=false` moves from the clone config to the checkout invocation, preserving the previous quietness.
- A failing ref still fails the install loudly through the existing `t_err` path (the whole sequence runs in the `if ! ( ... )` subshell).

### Testing

Verified against a local upstream repo exercising all three ref types — annotated tag, lightweight tag, branch: identical trees to the old clone form, no warnings; a nonexistent ref fails the install as before. `bash -n` and `shfmt -i 4 -sr -ci -s -d` (the repo's formatter settings) both clean.

**Disclosure per #1115/#1041:** prepared with AI assistance; the mechanism, all commands, and the resulting trees were verified by direct execution as described above.

Closes #1078


coolaj86:
Agreed. This needs to be fixed. Reviewing now.

coolaj86:
**Change** — `webi_git_clone` in `_webi/package-install.tpl.sh`: `git clone --branch <tag> --depth=1 --single-branch` → `git init` + `git fetch <tag> --depth=2` + `git checkout FETCH_HEAD`.

| | |
|---|---|
| Fixes | #1078 — `refs/tags/<tag> <sha> is not a commit!` from shallow clone of annotated tags |
| Affected | 14 `git_url` packages (serviceman, pg-essentials, 12 vim-*) |
| Commits | `e5ab9a5` (fix, back-compat init) + `34f8d8d` (depth=2 → updateable checkouts) |
| Rebase | onto `origin/main` `88be0ed`, FF-merged at `34f8d8d` |

**Tested**

| Test | Result |
|---|---|
| A/B tree equivalence (old vs new path, file:// upstream, annotated tag) | identical HEAD on git 2.20.1 / 2.34.1 / 2.47.3 |
| Old-git compat (git < 2.28) | `init -b main` rejected on 2.20.1; `git -c init.defaultBranch=main init` works (floor ~1.7.2, below old flow's 1.7.10) |
| Forward compat | git 2.47.3: full install via beta, no warnings, no defaultBranch hint |
| End-to-end via beta.webi.sh (Apple Container) | `serviceman` v1.0.1 + annotated `vim-commentary` v1.3 install clean, 0 warnings |
| Executed path (`sh -x` trace) | `init` / `fetch --depth=2` / `checkout FETCH_HEAD` ran; old clone path absent |
| Consecutive-release update fast-forward | `--depth=1` FAILED (unrelated histories); `--depth=2` OK |
| shellcheck / sh -n / CI | clean / clean / green |

Note: the warning reproduces only on some git versions/serving modes (not 2.20/2.34/2.47 in our runs) — the emitting path is removed entirely, so it is fixed wherever it manifests.

PR #1133, opened 2026-09-16
fix(webi): guard to stop `webi --init zsh` overriding user's preference

tomgidden:
If the user has already run `compinit`, re-running it is unnecessary, and overrides any configuration parameters used in that earlier invocation.  For example:

```sh
autoload -Uz compinit
compinit -C -d $HOME/.zcompdump_$ZSH_VERSION
```

In that example, the re-running of `compinit` without parameters by the `webi --init zsh` snippet at the end of `.zshrc` will cause the default `~/.zcompdump` to be created as well as the `~/.zcompdump_5.9` the user intended.

This commit adds a guard to check if `compdef` has been defined, which signals that `compinit` has already been run.


coolaj86:
Agreed, this needs to be fixed. Thank you.

